PR #2, opened 2021-06-09
sqlite: new package, database/sql driver

crawshaw:

### DIFF
--- a/sqlite.go
+++ b/sqlite.go
@@ -1,0 +1,558 @@
+// Copyright (c) 2021 Tailscale Inc & AUTHORS All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package sqlite implements a database/sql driver for SQLite3.
+//
+// This driver requires a file: URI always be used to open a database.
+// For details see https://sqlite.org/c3ref/open.html#urifilenames.
+//
+//
+// Memory Mode
+//
+// In-memory databases are popular for tests.
+// Use standard URIs to open in-memory databases, as documented
+// in the link above. E.g.
+//
+//	file:/whatever?mode=memory
+//
+// If opening a database in memory mode with the "?mode=memory" query
+// parameter, make sure immediately after opening to set
+//
+//	db.SetMaxOpenConns(1)
+//
+// This driver does not support shared cache mode, as the SQLite
+// authors recommend against it:
+//
+//	"Any use of shared cache is discouraged."
+//	https://www.sqlite.org/c3ref/enable_shared_cache.html
+//
+// For far more realistic in-memory database testing, use
+//
+//	PRAGMA journal_mode=WAL;
+//	PRAGMA synchronous=OFF;
+//
+// On any modern OS, this will generate syscalls as part of writing
+// data, but will never block on I/O.
+//
+//
+// Binding Types
+//
+// SQLite is flexible about type conversions, and so is this driver.
+// Almost all "basic" Go types (int, float64, string) are accepted and
+// directly mapped into SQLite, even if they are named Go types.
+// The time.Time type is also accepted (described below).
+// Values that implement encoding.TextMarshaler or json.Marshaler are
+// stored in SQLite in their marshaled form.
+//
+//
+// Binding Time
+//
+// While SQLite3 has no strict time datatype, it does have a series of built-in
+// functions that operate on timestamps that expect columns to be in one of many
+// formats: https://sqlite.org/lang_datefunc.html
+//
+// When encoding a time.Time into one of SQLite's preferred formats, we use the
+// shortest timestamp format that can accurately represent the time.Time.
+// The supported formats are:
+//
+//	2. YYYY-MM-DD HH:MM
+//	3. YYYY-MM-DD HH:MM:SS
+//	4. YYYY-MM-DD HH:MM:SS.SSS
+//
+// If the time.Time is not UTC (strongly consider storing times in UTC!),
+// we follow SQLite's norm of appending "[+-]HH:MM" to the above formats.
+//
+// It is common in SQLite to store "Unix time", seconds-since-epoch in an
+// INTEGER column. This is understood by the date and time functions documented
+// in the link above. If you want to do that, pass the result of time.Time.Unix
+// to the driver.
+//
+//
+// Reading Time
+//
+// In general, time is hard to extract from SQLite as a time.Time.
+// If a column is defined as DATE or DATETIME, then text data is parsed
+// as TimeFormat and returned as a time.Time. Integer data is parsed as
+// seconds since epoch and returned as a time.Time.
+package sqlite
+
+import (
+	"context"
+	"database/sql"
+	"database/sql/driver"
+	"encoding"
+	"errors"
+	"fmt"
+	"io"
+	"reflect"
+	"strings"
+	"time"
+
+	"github.com/tailscale/sqlite/cgosqlite"
+	"github.com/tailscale/sqlite/sqliteh"
+)
+
+// TimeFormat is the string format this driver uses to store
+// microsecond-precision time in SQLite in text format.
+const TimeFormat = "2006-01-02 15:04:05.000-0700"
+
+func init() {
+	sql.Register("sqlite3", drv{})
+}
+
+type drv struct{}
+
+func (d drv) Open(name string) (driver.Conn, error) { panic("deprecated, unused") }
+func (d drv) OpenConnector(name string) (driver.Connector, error) {
+	return &connector{name: name}, nil
+}
+
+type connector struct {
+	name string
+}
+
+func (c *connector) Driver() driver.Driver { return drv{} }
+func (c *connector) Connect(ctx context.Context) (driver.Conn, error) {
+	flags := sqliteh.SQLITE_OPEN_URI |
+		sqliteh.SQLITE_OPEN_READWRITE |
+		sqliteh.SQLITE_OPEN_CREATE
+	db, err := cgosqlite.Open(c.name, flags, "")
+	if err != nil {
+		e := &Error{
+			Code: sqliteh.Code(err.(sqliteh.ErrCode)),
+			Loc:  "Open",
+		}
+		if db != nil {
+			e.Msg = db.ErrMsg()
+			db.Close()
+		}
+		return nil, e
+	}
+	return &conn{db: db}, nil
+}
+
+type conn struct {
+	db    *cgosqlite.DB
+	stmts map[string]*stmt // persisted statements
+}
+
+func (c *conn) Prepare(query string) (driver.Stmt, error) { panic("deprecated, unused") }
+func (c *conn) Begin() (driver.Tx, error)                 { panic("deprecated, unused") }
+func (c *conn) Close() error {
+	return reserr(c.db, "Conn.Close", "", c.db.Close())
+}
+func (c *conn) PrepareContext(ctx context.Context, query string) (driver.Stmt, error) {
+	return c.prepare(query, false)
+}
+
+func (c *conn) prepare(query string, persist bool) (*stmt, error) {
+	if s := c.stmts[query]; s != nil {
+		return s, nil
+	}
+	cstmt, rem, err := c.db.Prepare(query, 0)
+	if err != nil {
+		return nil, reserr(c.db, "Prepare", query, err)
+	}
+	if rem != "" {
+		cstmt.Finalize()
+		return nil, &Error{
+			Code:  sqliteh.SQLITE_MISUSE,
+			Loc:   "Prepare",
+			Query: query,
+			Msg:   fmt.Sprintf("query has trailing text: %q", rem),
+		}
+	}
+	s := &stmt{db: c.db, stmt: cstmt, query: query, persist: persist}
+
+	if !persist {
+		return s, nil
+	}
+
+	if c.stmts == nil {
+		c.stmts = make(map[string]*stmt)
+	}
+	c.stmts[query] = s
+	return s, nil
+}
+
+func (c *conn) execInternal(ctx context.Context, query string) error {
+	s, err := c.prepare(query, true)
+	if err != nil {
+		if e, _ := err.(*Error); e != nil {
+			e.Loc = "internal:" + e.Loc
+		}
+		return err
+	}
+	if _, err := s.ExecContext(ctx, nil); err != nil {
+		return err
+	}
+	s.Close()
+	return nil
+}
+
+func (c *conn) BeginTx(ctx context.Context, opts driver.TxOptions) (driver.Tx, error) {
+	const LevelSerializable = 6 // matches the sql package constant
+	if opts.Isolation != 0 && opts.Isolation != LevelSerializable {
+		return nil, errors.New("github.com/tailscale/sqlite driver only supports serializable isolation level")
+	}
+	if opts.ReadOnly {
+		// TODO: match the state of the connection.
+		return nil, errors.New("github.com/tailscale/sqlite driver does not support read-only Tx yet")
+	}
+	if err := c.execInternal(ctx, "BEGIN"); err != nil {
+		return nil, err
+	}
+	return &connTx{c: c}, nil
+}
+
+type connTx struct {
+	c *conn
+}
+
+func (tx *connTx) Commit() error {
+	return tx.c.execInternal(context.Background(), "COMMIT")
+}
+
+func (tx *connTx) Rollback() error {
+	return tx.c.execInternal(context.Background(), "ROLLBACK")
+}
+
+func reserr(db *cgosqlite.DB, loc, query string, err error) error {
+	if err == nil {
+		return nil
+	}
+	e := &Error{
+		Code:  sqliteh.Code(err.(sqliteh.ErrCode)),
+		Loc:   loc,
+		Query: query,
+	}
+	// TODO(crawshaw): consider an API to expose this. sqlite.DebugErrMsg(db)?
+	if false {
+		e.Msg = db.ErrMsg()
+	}
+	return e
+}
+
+type stmt struct {
+	db      *cgosqlite.DB
+	stmt    *cgosqlite.Stmt
+	query   string
+	persist bool
+}
+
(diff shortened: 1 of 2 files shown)


sqlite.go
```diff
@@ -0,0 +1,542 @@
...
+// Package sqlite implements a database/sql driver for SQLite3.
+//
+// This driver requires a file: URI always be used to open a database.
+// For details see https://sqlite.org/c3ref/open.html#urifilenames.
+//
+//
+// Memory Mode
+//
+// In-memory databases are popular for tests.
+// If opening a database in memory mode with the "?mode=memory" query
```
danderson, line 14: Above it says to always use file://. How does this square with in-memory mode?

sqlite.go
```diff
@@ -0,0 +1,542 @@
...
+// data, but will never block on I/O.
+//
+//
+// Binding Types
+//
+// SQLite is flexible about type conversions, and so is this driver.
+// Almost all "basic" Go types (int, float64, string) are accepted and
+// directly mapped into SQLite, even if they are named Go types.
+// The time.Time type is also accepted (described below).
+// Unrecognized types are tested to be implementations of
```
danderson, line 40: "are tested to be" made my brain skip a gear and have to go back. How about:

Values that implement encoding.TextMarshaler or json.Marshaler are stored in SQLite in their marshaled form.

sqlite.go
```diff
@@ -0,0 +1,542 @@
...
+//
+// It is common for efficiency to store a time as seconds-since-epoch in an
+// INTEGER column. If you want to do that, pass the result of time.Time.Unix
+// to the driver.
+//
+//
+// Reading Time
+//
+// In general, time is hard to extract from SQLite as a time.Time.
+// If a column is defined as DATE or DATETIME, then the text can be parsed
```
danderson, line 69: "can be parsed"? Or "is parsed"?

sqlite.go
```diff
@@ -0,0 +1,542 @@
...
+const TimeFormat = "2006-01-02 15:04:05.000-0700"
+
+func init() {
+	sql.Register("sqlite3", drv)
+}
+
+// ErrMsg sets the driver to collect the sqlite3_errmsg string on error.
+// This is useful but generates extra string allocations.
+// Turn it off if you have written a program with an SQL error on a hot path.
+var ErrMsg = true
```
danderson, line 101: Can you make this an open option on the DB URL instead? It feels dirty to have mutable package-level variables.

sqlite.go
```diff
@@ -0,0 +1,542 @@
...
+
+var _ driver.Driver = Driver{}
```
danderson, line 103: Why this block?

sqlite.go
```diff
@@ -0,0 +1,542 @@
...
+var _ driver.DriverContext = Driver{}
+var _ driver.ConnBeginTx = (*conn)(nil)
+var _ driver.StmtExecContext = (*stmt)(nil)
+var _ driver.StmtQueryContext = (*stmt)(nil)
+
+var drv Driver
```
danderson, line 109: Why a random package-level driver?

sqlite.go
```diff
@@ -0,0 +1,542 @@
...
+
+type Driver struct{}
+
+func (d Driver) Open(name string) (driver.Conn, error) { panic("unused") }
```
danderson, line 113: Docstrings on exported types. You're welcome to add these later if you just want to hack non-blockingly for now.

danderson, line 113: Why both types? The database/sql/driver package suggests you should only implement driver.Connector and driver.DriverContext.

sqlite.go
```diff
@@ -0,0 +1,542 @@
...
+	return &connector{name: name}, nil
+}
+
+type connector struct {
+	name string
+}
+
+func (c *connector) Driver() driver.Driver { return drv }
+func (c *connector) Connect(ctx context.Context) (driver.Conn, error) {
+	log.Printf("connector.Connect: %s", c.name)
```
danderson, line 124: I generally view it rude for libraries to use the log package, because it leads to hacks such as the ones we've had to deploy for wireguard-go to capture them in other programs. Is this only for debugging?

sqlite.go
```diff
@@ -0,0 +1,542 @@
...
+	}
+	return &conn{db: db}, nil
+}
+
+type conn struct {
+	db    *cgosqlite.DB
+	stmts map[string]*stmt // persisted statements
+}
+
+func (c *conn) Prepare(query string) (driver.Stmt, error) { panic("unused") }
```
danderson, line 149: Should this delegate to prepare() also?

I'm going to assume all the panics in here will get implemented later.

sqlite.go
```diff
@@ -0,0 +1,542 @@
...
+		return err
+	}
+	s.Close()
+	return nil
+}
+
+func (c *conn) BeginTx(ctx context.Context, opts driver.TxOptions) (driver.Tx, error) {
+	const LevelSerializable = 6 // matches the sql package constant
+	if opts.Isolation != 0 && opts.Isolation != LevelSerializable {
+		return nil, errors.New("sqlite driver only supports serializable isolation level")
```
danderson, line 203: Does it not support linearizable as well? AFAICT sqlite's serializability translates to linearizability as well, but I'm bad at grokking edge cases.

sqlite_test.go
```diff
@@ -0,0 +1,252 @@
...
+		t.Fatal(err)
+	}
+	if _, err := db.Exec("PRAGMA synchronous=OFF"); err != nil {
+		t.Fatal(err)
+	}
+	return db
+}
+
+// execContexter is an *sql.DB or an *sql.Tx.
+type execContexter interface {
```
danderson, line 52: Could use driver.ExecerContext instead, which afaict is the same interface.

danderson (approved):

sqlite.go
crawshaw: `file:somefile?mode=memory` opens in-memory mode in SQLite.

sqlite.go
crawshaw: Useful during dev. Removed.

sqlite.go
crawshaw: unexported the type to save typing :)

crawshaw: Only Connect gets called, but `sql.Register` takes a `driver.Driver`, which has an Open method, so I need a dummy Open method.

sqlite.go
crawshaw: Debugging cruft. All gone.

sqlite.go
crawshaw: The `panic("unused")` are as above, methods that are no longer used but we need to meet the types. In this case, a `*conn` is a `driver.Conn` because that's what a `driver.Connector` returns, even though the sql package doesn't use the old methods, only the newer ones from `driver.ConnPrepareContext`.

I found navigating the database/sql/driver package types to be a very un-Go-like experience. :-/

sqlite.go
crawshaw: I think sqlite is linearizable, but I haven't proven it to myself. The sqlite docs explicitly call out serializable, so that's what I went with: https://sqlite.org/isolation.html

sqlite_test.go
crawshaw: That appears to return a `driver.Result`, not an `sql.Result`.

sqlite.go
```diff
@@ -0,0 +1,530 @@
...
+	case reflect.String:
+		_, err := s.bindBasic(debugName, arg.Ordinal, val.String())
+		return err
+	}
+
+	return &Error{
+		Code:  sqliteh.SQLITE_MISUSE,
+		Loc:   "Bind",
+		Query: s.query,
+		Msg:   fmt.Sprintf("Bind:%v: unknown value type %T", debugName, arg.Value),
```
josharian, line 369: Mention TextMarshaler here as a hint to the fix?

sqlite.go
```diff
@@ -0,0 +1,530 @@
...
+			Msg:   fmt.Sprintf("Bind:%v: sqlite does not support uint64", debugName),
+		}
+	case reflect.Uint8, reflect.Uint16, reflect.Uint32:
+		_, err := s.bindBasic(debugName, arg.Ordinal, int64(val.Uint()))
+		return err
+	case reflect.Float32, reflect.Float64:
+		_, err := s.bindBasic(debugName, arg.Ordinal, val.Float())
+		return err
+	case reflect.String:
+		_, err := s.bindBasic(debugName, arg.Ordinal, val.String())
```
josharian, line 361: It seems silly to allocate to put val.String() into an interface{} and then promptly type switch on the type to execute a single line of code and a defer. Ditto for most of the other calls to bindBasic. I suggest exploding bindBasic into separate methods, each of which you can call. Or inline here by adding defer call above and calling `s.stmt.Bind*` directly here.

sqlite.go
```diff
@@ -0,0 +1,530 @@
...
+
+	for i := range dest {
+		switch r.colDeclTypes[i] {
+		case "DATETIME", "DATE":
+			switch r.colTypes[i] {
+			case sqliteh.SQLITE_INTEGER:
+				v := r.stmt.stmt.ColumnInt64(i)
+				dest[i] = time.Unix(v, 0)
+			case sqliteh.SQLITE_FLOAT:
+				dest[i] = r.stmt.stmt.ColumnDouble(i)
```
josharian, line 477: There's nothing documented about storing floats in date/datetime columns, but it seems (naively) to me like we should treat it equivalently to ints, that is, time.Unix, with the fractional part being fractions of seconds.

Or treat it as an error.

sqlite.go
```diff
@@ -0,0 +1,530 @@
...
+			case sqliteh.SQLITE_TEXT:
+				v := r.stmt.stmt.ColumnText(i)
+				t, err := time.Parse(TimeFormat, v)
```
josharian, line 480: time.Parse will handle truncated inputs correctly? That's neat.

sqlite.go
```diff
@@ -0,0 +1,530 @@
...
+					return fmt.Errorf("cannot parse time from column %d: %v", i, err)
+				}
+				dest[i] = t
+			}
+			continue
+		}
+		switch r.colTypes[i] {
+		case sqliteh.SQLITE_INTEGER:
+			val := r.stmt.stmt.ColumnInt64(i)
+			if strings.EqualFold(r.colDeclTypes[i], "BOOLEAN") {
```
josharian, line 491: Why is this EqualFold, but the main switch statement above isn't?

sqlite.go
```diff
@@ -0,0 +1,530 @@
...
+// Error is an error produced by SQLite.
+type Error struct {
+	Code  sqliteh.Code // SQLite extended error code (SQLITE_OK is an invalid value)
+	Loc   string       // method name that generated the error
+	Query string       // original SQL query text
+	Msg   string       // value of sqlite3_errmsg, set sqlite.ErrMsg = true
+}
+
+func (err Error) Error() string {
+	str := "sqlite"
```
josharian, line 518: If you're concerned about allocating extra times during the error path (as above, see ErrMsg), which I approve of, then should probably use a strings.Builder here.

sqlite_test.go
```diff
@@ -0,0 +1,252 @@
...
+	if want := t1Str; txt != want {
+		t.Errorf("time stored as %q, want %q", txt, want)
+	}
+
+	exec(t, db, "CREATE TABLE t2 (c FOOD)")
+	exec(t, db, "INSERT INTO t2 VALUES (?)", t1)
+	if err := db.QueryRowContext(context.Background(), "SELECT c FROM t2").Scan(&t2); err == nil {
+		t.Fatal("expect an error trying to interpet FOOD as Time")
+	}
+}
```
josharian, line 252: Test TextMarshal?

Benchmarks?

sqlite_test.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+	if string(cBlob) != "txt" {
+		t.Errorf("cBlob=%v, want 'txt'", cBlob)
+	}
+	if cNull != nil {
+		t.Errorf("cNull=%v, want nil", cNull)
+	}
+}
+
+func TestTime(t *testing.T) {
```
josharian, line 223: Test the various shortened time formats?

sqlite.go
Xe: would it be good to clone how upstream sqlite3 does it? that uses `:memory:`

crawshaw: This is how sqlite3 does it. The ":memory:" mode is their pre-URI naming scheme that doesn't let you distinguish different names. In this form, file:whatever?mode=memory is an in-memory table named whatever.

Examples and details are in the SQLite docs linked a few lines earlier. I'm reluctant to repeat their docs here much. I put one example in.

sqlite.go
crawshaw: I haven't got into the business of parsing the URL yet and I would like to avoid it. I'll investigate an API for this later. For now I've removed it.

sqlite.go
```diff
@@ -0,0 +1,530 @@
...
+// For details see https://sqlite.org/c3ref/open.html#urifilenames.
+//
+//
+// Memory Mode
+//
+// In-memory databases are popular for tests.
+// If opening a database in memory mode with the "?mode=memory" query
+// parameter, make sure immediatley after opening to set
+//
+//	db.SetMaxOpenConns(1)
```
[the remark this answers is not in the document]
crawshaw: I can't. SetMaxOpenConns is a method on the sql.DB object that the driver never sees.

sqlite.go
```diff
@@ -0,0 +1,530 @@
...
+// shortest timestamp that the time.Time can be represented using, one of:
+//
+//	2. YYYY-MM-DD HH:MM
+//	3. YYYY-MM-DD HH:MM:SS
+//	4. YYYY-MM-DD HH:MM:SS.SSS
+//
+// If the time.Time is not UTC (strongly consider storing times in UTC!),
+// we follow SQLite's norm of appending "[+-]HH:MM" to the above formats.
+//
+// It is common for efficiency to store a time as seconds-since-epoch in an
```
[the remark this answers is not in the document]
crawshaw: Because that's how SQLite does it. Added another comment pointing to the URL above.

sqlite.go
```diff
@@ -0,0 +1,530 @@
...
+// microsecond-precision time in SQLite in text format.
+const TimeFormat = "2006-01-02 15:04:05.000-0700"
+
+func init() {
+	sql.Register("sqlite3", drv{})
+}
+
+// ErrMsg sets the driver to collect the sqlite3_errmsg string on error.
+// This is useful but generates extra string allocations.
+// Turn it off if you have written a program with an SQL error on a hot path.
```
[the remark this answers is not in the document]
crawshaw: It's gone.

sqlite.go
```diff
@@ -0,0 +1,530 @@
...
+}
+func (s *stmt) Exec(args []driver.Value) (driver.Result, error) { panic("unused") }
+func (s *stmt) Query(args []driver.Value) (driver.Rows, error)  { panic("unused") }
+
+func (s *stmt) ExecContext(ctx context.Context, args []driver.NamedValue) (driver.Result, error) {
+	if err := s.stmt.Reset(); err != nil {
+		return nil, s.reserr("Stmt.Query(Reset)", err)
+	}
+	defer s.stmt.ClearBindings()
+	defer s.stmt.Reset()
```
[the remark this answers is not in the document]
crawshaw: It is documented to only return whatever error sqlite3_step returned. Commented to say that.

sqlite.go
```diff
@@ -0,0 +1,530 @@
...
+	for _, arg := range args {
+		if err := s.bind(arg); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (s *stmt) bind(arg driver.NamedValue) error {
+	var debugName interface{} = arg.Ordinal
```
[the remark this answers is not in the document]
crawshaw: Done and added a TODO.

sqlite.go
```diff
@@ -0,0 +1,530 @@
...
+		b := int64(0)
+		if val.Bool() {
+			b = 1
+		}
+		_, err := s.bindBasic(debugName, arg.Ordinal, b)
+		return err
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		_, err := s.bindBasic(debugName, arg.Ordinal, val.Int())
+		return err
+	case reflect.Uint, reflect.Uint64:
```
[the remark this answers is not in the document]
crawshaw: They can? This code is handling when someone passes in a uint64 directly. If they pass in a uint64 type which is a TextMarshaler, it's caught just above and handled. They can also make their own string and pass it in.

Not really sure what you want me to do here.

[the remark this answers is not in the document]
josharian: Oh, I see. I missed the order of operations here, because I was flipping back and forth to bindBasic a lot. I still am tempted to try to make uint and uint64 just work, at least sometimes, but I guess that's a lost cause.

sqlite.go
crawshaw: I agree. I spent a bit of time trying to fix this without making a mess of the errors. I think it needs more tests before I can do it safely.

sqlite.go
crawshaw: Added as TODO

sqlite.go
crawshaw: So I'm sure I tested this. But apparently it doesn't work? Oops.

sqlite_test.go
crawshaw: Added some TODOs.

sqlite.go
```diff
@@ -0,0 +1,530 @@
...
+				Code:  sqliteh.SQLITE_MISUSE,
+				Loc:   "Bind",
+				Query: s.query,
+				Msg:   fmt.Sprintf("unknown parameter name %q", arg.Name),
+			}
+		}
+		arg.Ordinal = index
+	}
+
+	// Start with obvious types, including time.Time before TextMarshaler.
```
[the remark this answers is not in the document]
crawshaw: I believe this code does the right thing, it tests the direct concrete types, which will not catch `type T int` or its ilk. Then TextMarshaler, then it tries to extract the basic type using reflect.

But I'll write more tests in a followup.

PR #4, opened 2021-07-08
sqlite: add WithPersist

crawshaw:

### DIFF
--- a/sqlite.go
+++ b/sqlite.go
@@ -143,14 +143,20 @@ func (c *conn) Close() error {
 	return reserr(c.db, "Conn.Close", "", c.db.Close())
 }
 func (c *conn) PrepareContext(ctx context.Context, query string) (driver.Stmt, error) {
-	return c.prepare(query, false)
+	persist := ctx.Value(persistQuery{}) != nil
+	return c.prepare(query, persist)
 }
 
 func (c *conn) prepare(query string, persist bool) (*stmt, error) {
+	query = strings.TrimSpace(query)
 	if s := c.stmts[query]; s != nil {
 		return s, nil
 	}
-	cstmt, rem, err := c.db.Prepare(query, 0)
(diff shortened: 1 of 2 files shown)


sqlite.go
```diff
@@ -556,3 +562,13 @@ func (err Error) Error() string {
...
+
+// WithPersist makes a ctx instruct the sqlite driver to persist a prepared query.
+//
+// This should be used with recurring queries to avoid constant parsing and
+// planning of the query by SQLite.
+func WithPersist(ctx context.Context) context.Context {
+	return context.WithValue(ctx, persistQuery(0), persistQuery(0))
+}
+
+type persistQuery int
```
bradfitz, line 574: use an empty struct to avoid boxing into an interface{}

and document that it's a context value

crawshaw: done.

sqlite.go
```diff
@@ -556,3 +562,14 @@ func (err Error) Error() string {
 	}
 	return b.String()
 }
+
+// WithPersist makes a ctx instruct the sqlite driver to persist a prepared query.
```
josharian, line 566: It's unfortunate to have to make a new context for this, because it allocates. It forces callers to allocate each time or maintain a separate "first-time-through" code path. I don't have any answers, just wanted to call it out.

josharian (approved):

sqlite.go
crawshaw: There are so many allocations through the use of database/sql it gets drowned in the noise. :-/

FWIW, a really enthusiastic user could `sqlite.WithPersist` all their queries once, and then not do that in the future. Though that would force a lot of the accounting work on the user I'm trying to avoid. I'll ponder workarounds. One would be a `func DefaultPersist(conn)` but that's surprisingly hard to use with database/sql's pool.

PR #5, opened 2021-07-08
sqlite: add ExecScript

crawshaw:

### DIFF
--- a/sqlite.go
+++ b/sqlite.go
@@ -556,3 +556,49 @@ func (err Error) Error() string {
 	}
 	return b.String()
 }
+
+// SQLConn is a database/sql.Conn.
+// (We cannot create a circular package dependency here.)
+type SQLConn interface {
+	Raw(func(driverConn interface{}) error) error
+}
+
+// ExecScript executes a set of SQL queries on an sql.Conn.
+// It stops on the first error.
+// It is recommended you wrap your script in a BEGIN; ... COMMIT; block.
+//
+// Usage:
+//
+//	c, err := db.Conn(ctx)
+//	if err != nil {
+//		// handle err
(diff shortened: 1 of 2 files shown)


sqlite.go
```diff
@@ -556,3 +556,45 @@ func (err Error) Error() string {
...
+//	if err != nil {
+//		// handle err
+//	}
+//	if err := sqlite.ExecScript(c, queries); err != nil {
+//		// handle err
+//	}
+//	c.Close() // return sql.Conn to pool
+func ExecScript(sqlconn SQLConn, queries string) error {
+	return sqlconn.Raw(func(driverConn interface{}) error {
+		c := driverConn.(*conn)
```
josharian, line 580: Do comma-ok here and return an error if !ok. Nicer than panicking.

sqlite.go
```diff
@@ -556,3 +556,45 @@ func (err Error) Error() string {
...
+
+		for {
+			queries = strings.TrimSpace(queries)
+			if queries == "" {
+				break
```
josharian, line 585: Make this `return nil`? And then remove the `return nil` on line 598.

sqlite.go
```diff
@@ -556,3 +556,45 @@ func (err Error) Error() string {
...
 	return b.String()
 }
+
+// SQLConn is a database/sql.Conn.
+// (We cannot create a circular package dependency here.)
+type SQLConn interface {
+	Raw(func(driverConn interface{}) error) error
+}
+
+// ExecScript executes a set of SQL queries on an sql.Conn.
```
josharian, line 566: document that it stops on error

josharian (approved):

PR #6, opened 2021-07-10
sqlite: speed up persistent queries

crawshaw:
Best read commit-by-commit. Each includes relative benchmark performance to the commit before. Overall results of this PR:

```
name        old time/op  new time/op  delta
Persist-24  2.68µs ± 1%  2.03µs ± 1%  -24.00%  (p=0.000 n=20+20)
```

### DIFF
--- a/sqlite_test.go
+++ b/sqlite_test.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"runtime"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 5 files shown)


sqlite_test.go
```diff
@@ -34,7 +34,7 @@ func TestOpenDB(t *testing.T) {
 	}
 }
 
-func openTestDB(t *testing.T) *sql.DB {
+func openTestDB(t testing.TB) *sql.DB {
```
josharian, line 38: Someday I'll learn to write it with testing.TB from the beginning. Someday.

josharian (approved):
Nice.

PR #9, opened 2021-07-14
sqlite, etc: add tracing API

crawshaw:
Our expected use of this will be as an "event" API, not a tracing API.
But for future users, I included a traceID, that could be used to
retrofit one of those large trace/span systems onto the driver if you
were so inclined.

Minimal benchmark impact:

	Persist-24    1.93µs ± 1%  1.92µs ± 1%  -0.51%  (p=0.000 n=19+18)
	EmptyExec-24   386ns ± 0%   401ns ± 1%  +3.85%  (p=0.000 n=17+20)

### DIFF
--- a/cgosqlite/cgosqlite.go
+++ b/cgosqlite/cgosqlite.go
@@ -27,7 +27,7 @@ package cgosqlite
 // #cgo CFLAGS: -DSQLITE_ENABLE_COLUMN_METADATA
 // #cgo CFLAGS: -DSQLITE_ENABLE_STAT4
 // #cgo CFLAGS: -DHAVE_USLEEP=1
-// #cgo linux LDFLAGS: -ldl -lm
+// #cgo linux LDFLAGS: -ldl -lm -lrt
 // #cgo linux CFLAGS: -std=c99
 //
 // #include <stdint.h>
@@ -35,6 +35,7 @@ package cgosqlite
 // #include <string.h>
 // #include <pthread.h>
 // #include <sqlite3.h>
+// #include <time.h>
 // #include "cgosqlite.h"
 import "C"
 import (
@@ -57,8 +58,13 @@ type DB struct {
 
 // Stmt implements sqliteh.Stmt.
 type Stmt struct {
-	db   *DB
-	stmt *C.sqlite3_stmt
+	db    *DB
+	stmt  *C.sqlite3_stmt
+	start C.struct_timespec
+
+	// used as scratch space when calling into cgo
+	rowid, changes C.sqlite3_int64
+	duration       C.int64_t
 }
 
 // Open implements sqliteh.OpenFunc.
@@ -153,8 +159,17 @@ func (stmt *Stmt) ClearBindings() error {
 	return errCode(C.sqlite3_clear_bindings(stmt.stmt))
 }
 
-func (stmt *Stmt) ResetAndClear() error {
-	return errCode(C.reset_and_clear(stmt.stmt))
+func (stmt *Stmt) ResetAndClear() (time.Duration, error) {
+	if stmt.start != (C.struct_timespec{}) {
+		stmt.duration = 0
+		err := errCode(C.reset_and_clear(stmt.stmt, &stmt.start, &stmt.duration))
+		return time.Duration(stmt.duration), err
+	}
+	return 0, errCode(C.reset_and_clear(stmt.stmt, nil, nil))
+}
+
+func (stmt *Stmt) StartTimer() {
+	C.monotonic_clock_gettime(&stmt.start)
 }
 
 func (stmt *Stmt) ColumnDatabaseName(col int) string {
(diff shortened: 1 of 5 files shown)


cgosqlite/cgosqlite.go
```diff
@@ -148,8 +151,17 @@ func (stmt *Stmt) ClearBindings() error {
 	return errCode(C.sqlite3_clear_bindings(stmt.stmt))
 }
 
-func (stmt *Stmt) ResetAndClear() error {
-	return errCode(C.reset_and_clear(stmt.stmt))
+func (stmt *Stmt) ResetAndClear() (time.Duration, error) {
+	if stmt.start != (C.struct_timespec{}) {
+		var d C.int64_t
+		err := errCode(C.reset_and_clear(stmt.stmt, &stmt.start, &d))
```
josharian, line 157: I think this'll force d onto the stack. Do you have any benchmarks and/or alloc tests that cover this? You may want to return a struct from reset_and_clear instead. (cgo is so awkward.)

cgosqlite/cgosqlite.go
```diff
@@ -172,19 +184,21 @@ func (stmt *Stmt) Step() (row bool, err error) {
 	}
 }
 
-func (stmt *Stmt) StepResult() (row bool, lastInsertRowID, changes int64, err error) {
+func (stmt *Stmt) StepResult() (row bool, lastInsertRowID, changes int64, d time.Duration, err error) {
 	var rowid, chng C.sqlite3_int64
-	res := C.step_result(stmt.stmt, &rowid, &chng)
+	var duration C.int64_t
+	res := C.step_result(stmt.stmt, &rowid, &chng, &duration)
```
josharian, line 190: Ditto: I think duration will be heap-allocated here.

One option here if you don't want to switch to returning a giant C struct is to heap allocate a single larger type:

```go
x := new(struct{
  rowid, chng C.sqlite3_int64
  duration C.int64_t
})
res := C.step_result(stmt.stmt, &x.rowid, &x.chng, &x.duration)
```

But better would be to return a C struct. (All this conditional on tests/benchmarks of the allocs here showing that that my mental escape analysis is correct.)

cgosqlite/cgosqlite.h
```diff
@@ -28,21 +28,50 @@ static int bind_parameter_index(sqlite3_stmt* stmt, _GoString_ s) {
...
+	// defined properly here by cgo. My C compiler seems otherwise
+	// fine, I can compile a small C program on linux referring to
+	// CLOCK_MONOTONIC.
+	// 
+	// For now, use the value directly.
+	// It is defined in POSIX so it won't change any time soon.
+	clock_gettime(1, t); // CLOCK_MONOTONIC
+}
+
+static int64_t time_since_ns(const struct timespec t1)
```
josharian, line 42: Suggest naming this `ns_since`

sqlite.go
```diff
@@ -96,6 +96,12 @@ var Open sqliteh.OpenFunc = func(string, sqliteh.OpenFlags, string) (sqliteh.DB,
 	return nil, fmt.Errorf("cgosqlite.Open is missing")
 }
 
+// TraceFunc is function called by the driver to report a completed query.
```
josharian, line 99: This doc could use a bit more detail.

A reader might wonder: Does a completed query's duration include the time needed to step over each of the results? If so, is it just the time spent talking to the DB, accumulated step-by-step, or it is the total time elapsed from starting query to last result fetched?

Is the query string potentially sensitive, or is it the version with `?`s in it (which could presumably be safely logged)?

sqlite_test.go
```diff
@@ -361,6 +372,95 @@ func TestWithPersist(t *testing.T) {
...
+		if !ok {
+			t.Fatal("trace: no event")
+		}
+		if ev.traceID != traceID {
+			t.Errorf("trace: traceID=%d, want %d", traceID, ev.traceID)
+		}
+		if ev.query != query {
+			t.Errorf("trace: query=%q, want %q", ev.query, query)
+		}
+		if ev.err == nil && errSubstr != "" {
```
josharian, line 398: Use a `switch {` statement?

sqlite_test.go
```diff
@@ -361,6 +372,95 @@ func TestWithPersist(t *testing.T) {
...
+			t.Errorf("trace: err=nil, want %q", errSubstr)
+		} else if ev.err != nil && errSubstr == "" {
+			t.Errorf("trace: err=%v, want nil", ev.err)
+		} else if ev.err != nil && !strings.Contains(ev.err.Error(), errSubstr) {
+			t.Errorf("trace: err=%v, want %v", ev.err, errSubstr)
+		}
+		if ev.duration == 0 || ev.duration > 10*time.Minute {
```
josharian, line 405: `<= 0`?

sqliteh/sqliteh.go
```diff
@@ -82,7 +86,7 @@ type Stmt interface {
 	// 	For SQLITE_DONE, Step returns row=false err=nil
 	// 	For any error, Step returns row=false err.
 	// https://www.sqlite.org/c3ref/step.html
-	StepResult() (row bool, lastInsertRowID, changes int64, err error)
+	StepResult() (row bool, lastInsertRowID, changes int64, _ time.Duration, err error)
```
josharian, line 89: Why `_` here instead of (say) `d`?

sqliteh/sqliteh.go
```diff
@@ -55,11 +56,14 @@ type Stmt interface {
...
 	// https://www.sqlite.org/c3ref/expanded_sql.html
 	ExpandedSQL() string
+	// StartTimer starts recording elapsed duration.
+	StartTimer()
 	// Reset is sqlite3_reset.
 	// https://www.sqlite.org/c3ref/reset.html
 	Reset() error
 	// ResetAndClear is sqlite3_reset + sqlite3_clear_bindings.
-	ResetAndClear() error
+	// It reports the duration passed since the call to StartTimer.
```
josharian, line 65: s/passed/elapsed/

josharian (approved):

cgosqlite/cgosqlite.go
crawshaw: Fixed by cheating, I put some scratch variables inside the *Stmt object that is already heap allocated. (Use of it is single-threaded.)

PR #13, opened 2021-09-25
all: export Checkpoint and BusyTimeout

crawshaw:

### DIFF
--- a/cgosqlite/cgosqlite.go
+++ b/cgosqlite/cgosqlite.go
@@ -118,6 +118,19 @@ func (db *DB) BusyTimeout(d time.Duration) {
 	C.sqlite3_busy_timeout(db.db, C.int(d/1e6))
 }
 
+func (db *DB) Checkpoint(dbName string, mode sqliteh.Checkpoint) (int, int, error) {
+	var cDB *C.char
+	if dbName != "" {
+		// Docs say: "If parameter zDb is NULL or points to a zero length string",
+		// so they are equivalent here.
+		cDB = C.CString(dbName)
+		defer C.free(unsafe.Pointer(cDB))
+	}
+	var nLog, nCkpt C.int
+	res := C.sqlite3_wal_checkpoint_v2(db.db, cDB, C.int(mode), &nLog, &nCkpt)
+	return int(nLog), int(nCkpt), errCode(res)
+}
+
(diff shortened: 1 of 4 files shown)


crawshaw:
TBR. Will do fixups later.

sqlite.go
```diff
@@ -695,12 +695,41 @@ func ExecScript(sqlconn SQLConn, queries string) error {
...
+		if !ok {
+			return fmt.Errorf("sqlite.BusyTimeout: sql.Conn is not the sqlite driver: %T", driverConn)
+		}
+		c.db.BusyTimeout(d)
+		return nil
+	})
+}
+
+// Checkpoint calls sqlite3_wal_checkpoint_v2 on the underlying connection.
+func Checkpoint(sqlconn SQLConn, dbName string, mode sqliteh.Checkpoint) (numFrames, numFramesCheckpointed int, err error) {
```
josharian, line 721: Maybe call this Checkpoint2 or CheckpointV2?

sqlite.go
```diff
@@ -695,12 +695,41 @@ func ExecScript(sqlconn SQLConn, queries string) error {
 			_, err = cstmt.Step()
 			cstmt.Finalize()
 			if err != nil {
+				// TODO(crawshaw): consider checking sqlite3_txn_state
+				// here and issuing a rollback, incase this script was:
```
josharian, line 699: in case

cgosqlite/cgosqlite.go
```diff
@@ -118,6 +118,19 @@ func (db *DB) BusyTimeout(d time.Duration) {
 	C.sqlite3_busy_timeout(db.db, C.int(d/1e6))
 }
 
+func (db *DB) Checkpoint(dbName string, mode sqliteh.Checkpoint) (int, int, error) {
+	var cDB *C.char
+	if dbName != "" {
+		// Docs say: "If parameter zDb is NULL or points to a zero length string",
+		// so they are equivalent here.
+		cDB = C.CString(dbName)
+		defer C.free(unsafe.Pointer(cDB))
```
josharian, line 127: `defer` in an if statement (instead of top level) is more expensive. Seems like checkpointing is expensive enough that that doesn't matter, but thought I should mention it just in case.

josharian (commented):
LGTM, with comments

josharian:
Cleaning out my inbox. Gentle reminder to do follow-up as appropriate.

PR #14, opened 2021-09-26
all: add TxnState

crawshaw:
Part of a debugging attempt.

### DIFF
--- a/cgosqlite/cgosqlite.go
+++ b/cgosqlite/cgosqlite.go
@@ -131,6 +131,15 @@ func (db *DB) Checkpoint(dbName string, mode sqliteh.Checkpoint) (int, int, erro
 	return int(nLog), int(nCkpt), errCode(res)
 }
 
+func (db *DB) TxnState(schema string) sqliteh.TxnState {
+	var cSchema *C.char
+	if schema != "" {
+		cSchema = C.CString(schema)
+		defer C.free(unsafe.Pointer(cSchema))
+	}
+	return sqliteh.TxnState(C.sqlite3_txn_state(db.db, cSchema))
+}
+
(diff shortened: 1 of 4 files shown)


sqliteh/sqliteh.go
```diff
@@ -354,6 +356,32 @@ func (mode Checkpoint) String() string {
 	}
 }
 
+// Txn is a transaction state.
+// It is used by sqlite3_txn_state.
+//
+// https://sqlite.org/c3ref/txn_state.html
+type Txn int
```
bradfitz, line 363: TxnState ?

sqlite.go
```diff
@@ -717,6 +717,18 @@ func BusyTimeout(sqlconn SQLConn, d time.Duration) error {
 	})
 }
 
+// TxnState calls sqlite3_txn_state on the underlying connection.
+func TxnState(sqlconn SQLConn, schema string) (state sqliteh.Txn, err error) {
```
bradfitz, line 721: if you rename the type, then you can drop the named result parameters.

the func name would need to change, though.... `ConnTxnState` ?

cgosqlite/cgosqlite.go
```diff
@@ -131,6 +131,15 @@ func (db *DB) Checkpoint(dbName string, mode sqliteh.Checkpoint) (int, int, erro
 	return int(nLog), int(nCkpt), errCode(res)
 }
 
+func (db *DB) TxnState(schema string) sqliteh.Txn {
```
bradfitz, line 134: docs? at least to clarify that schema is optional?

bradfitz (approved):

cgosqlite/cgosqlite.go
crawshaw: These methods are documented on the interface they implement. Those link to the sqlite docs, because I don't want to reproduce those here.

PR #15, opened 2021-09-29
sqlite: conn init mechanism

crawshaw:

### DIFF
--- a/sqlite.go
+++ b/sqlite.go
@@ -8,6 +8,23 @@
 // For details see https://sqlite.org/c3ref/open.html#urifilenames.
 //
 //
+// Initializing connections or tracing
+//
+// If you want to do initial configuration of a connection, or enable
+// tracing, use the Connector function:
+//
+//	connInitFunc := func(ctx context.Context, conn driver.ConnPrepareContext) error {
+//		called++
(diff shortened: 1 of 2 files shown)


sqlite.go
```diff
@@ -8,6 +8,24 @@
 // For details see https://sqlite.org/c3ref/open.html#urifilenames.
 //
 //
+// Initializing connections or tracing
+//
+// The database/sql package creates connections unexpectedly, so if you
```
josharian, line 13: I'm not sure this is quite the right messaging. :P

I'd cut the first part entirely.

```go
// If you want to do initial [...]
```

sqlite.go
```diff
@@ -93,6 +111,8 @@ var Open sqliteh.OpenFunc = func(string, sqliteh.OpenFlags, string) (sqliteh.DB,
 // code between calls to rows.Next.
 type TraceFunc func(prepCtx context.Context, query string, duration time.Duration, err error)
 
+type ConnInitFunc func(ctx context.Context, conn driver.ConnPrepareContext) error
```
josharian, line 114: Document what state conn is in when a ConnInitFunc gets called. What things can you / can you not do in this function? What happens if you return an error?

PR #17, opened 2021-10-29
sqlite: implement backup api

crawshaw:

### DIFF
--- a/examples/backup/backup.go
+++ b/examples/backup/backup.go
@@ -1,0 +1,181 @@
+package main
+
+import (
+	"context"
+	"database/sql"
+	"database/sql/driver"
+	"log"
+	"runtime"
+	"sync"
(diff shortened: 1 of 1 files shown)


raggi:
This is not safe as-is, sqlite itself does not lock results, which the backup API also uses. We will need to add a lock shared between the backup context and the connection context that serializes calls to step and query/result functions. I tested THREADSAFE=1 to see if that would do the work for us, it does not.

raggi:
TODO: test if we can perform online backups of two schemas concurrently

At time of writing control has two databases attached in the writable connection, main and noise. Ideally to backup both efficiently using the online backup API we'd be able to start a backup of both of these schemas and Step() them concurrently, and Finish() them at the same time, so the backups are in sync.

PR #20, opened 2021-11-28
cgosqlite: import SQLite amalgamation v3.37.0

crawshaw:


danderson (approved):
Hah, I had a commit for this locally and straight up forgot to send it. Thanks!

PR #24, opened 2021-12-31
sqlite: add CopyAll / DropAll helper functions

crawshaw:


util.go
```diff
@@ -0,0 +1,146 @@
+package sqlite
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"strings"
+)
+
+// DropAll deletes all the data from a database.
```
bradfitz, line 10: document schemaName and its default value

util.go
```diff
@@ -0,0 +1,146 @@
...
+			return fmt.Errorf("unknown sqlite schema type %q for %q", sqlType, name)
+		}
+	}
+	rows.Close()
+	if err := rows.Err(); err != nil {
+		return err
+	}
+
+	for _, name := range indexes {
+		if _, err := conn.ExecContext(ctx, fmt.Sprintf("DROP INDEX %q.%q", schemaName, name)); err != nil {
```
bradfitz, line 53: I guess %q is close enough?

util.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// CopyAll copies the contents of one database to another.
+//
+// Traditionally this is done in sqlite by closing the database and copying
+// the file. However it can be useful to do it online: a single exclusive
+// transaction can cross multiple databases, and if multiple processes are
+// using a file, this lets one replace the database without first
+// communicating with the other processes, asking them to close the DB first.
+func CopyAll(ctx context.Context, conn *sql.Conn, dstSchemaName, srcSchemaName string) (err error) {
```
bradfitz, line 82: Document empty schema default

bradfitz (approved):

util.go
crawshaw: surprisingly close. I believe %q can be used to generate invalid schema and table names, but the escaping appears correct for the set of names it can generate.

util.go
crawshaw: documented that this matches sqlite's pragma schema-name

PR #25, opened 2022-01-01
sqlite: CopyAll/DropAll: ignore automatic indexes

crawshaw:


util.go
```diff
@@ -24,7 +24,7 @@ func DropAll(ctx context.Context, conn *sql.Conn, schemaName string) (err error)
 
 	var indexes, tables, triggers, views []string
 
-	rows, err := conn.QueryContext(ctx, fmt.Sprintf("SELECT name, type FROM %q.sqlite_schema", schemaName))
+	rows, err := conn.QueryContext(ctx, fmt.Sprintf("SELECT name, type FROM %q.sqlite_schema WHERE sql != ''", schemaName))
```
bradfitz, line 27: TIL that `!=` even works at all. I thought you had to use `<>`. Do we have a style preference one way or another?

bradfitz: Add a comment to save a future from researching what this means...

> https://www.sqlite.org/schematab.html
> ...
> The sqlite_schema.sql is NULL for the internal indexes that are automatically created by UNIQUE or PRIMARY KEY constraints.

bradfitz (approved):

util.go
crawshaw: Depends on the SQL dialect. There are so many to choose from! But for sqlite: `The non-equals operator can be either != or <>. `

PR #29, opened 2022-01-27
sqlite, cgosqlite: fix TestTrace on macOS

crawshaw:
--- FAIL: TestTrace (0.00s)
	    sqlite_test.go:522: trace: improbable duration: -513288h42m26.15796988s
	    sqlite_test.go:523: trace: improbable duration: -513288h42m26.15796988s
	    sqlite_test.go:558: trace: improbable duration: 0s

Two problems:

1. It turns out CLOCK_MONOTONIC is 6 on macOS. Oops.
   My comment claims it was defined as 1 in the POSIX spec,
   but now I can't find it. So I can only say it's 1 on linux.

2. clock_gettime does not seem to have much resolution on
   macOS, so I see some zeros. Ignore them.

### DIFF
--- a/cgosqlite/cgosqlite.go
+++ b/cgosqlite/cgosqlite.go
@@ -31,6 +31,14 @@ package cgosqlite
 // #cgo linux LDFLAGS: -ldl -lm -lrt
 // #cgo linux CFLAGS: -std=c99
 //
+// // TODO(crawshaw): for some reason, I cannot get CLOCK_MONOTONIC
+// // defined properly here by cgo. My C compiler seems otherwise
+// // fine, I can compile a small C program on linux referring to
+// // CLOCK_MONOTONIC.
+// //
+// // For now, use the value directly. It's a fixed value on linux.
+// #cgo linux CFLAGS: -DCLOCK_MONOTONIC=1
+//
 // #include <stdint.h>
 // #include <stdlib.h>
 // #include <string.h>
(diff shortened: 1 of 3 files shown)


cgosqlite/cgosqlite.go
```diff
@@ -31,6 +31,14 @@ package cgosqlite
 // #cgo linux LDFLAGS: -ldl -lm -lrt
 // #cgo linux CFLAGS: -std=c99
 //
+// // TODO(crawshaw): for some reason, I cannot get CLOCK_MONOTONIC
+// // defined properly here by cgo. My C compiler seems otherwise
+// // fine, I can compile a small C program on linux referring to
+// // CLOCK_MONOTONIC.
+// //
+// // For now, use the value directly. It's fixed on linux.
```
bradfitz, line 39: what does "It's fixed on linux" mean?

oh! as in "it's a fixed value on linux".

I thought there was a bug being referred to.

sqlite_test.go
```diff
@@ -514,7 +514,11 @@ func TestTrace(t *testing.T) {
 			t.Errorf("trace: err=%v, want %v", ev.err, errSubstr)
 		}
 		if ev.duration <= 0 || ev.duration > 10*time.Minute {
-			t.Errorf("trace: improbable duration: %v", ev.duration)
+			// The macOS clock appears to low resolution and so
+			// it's common to get a duration of exactly 0s.
+			if runtime.GOOS == "darwin" && ev.duration != 0 {
```
bradfitz, line 519: shouldn't the first part be GOOS != darwin ?

Otherwise this error t.Errorf goes away on Linux.

bradfitz (approved):

sqlite_test.go
bradfitz: You marked this as resolved but I still don't understand it.

It looks wrong still.

crawshaw: I forgot to git add the file: https://github.com/tailscale/sqlite/pull/30

PR #33, opened 2022-03-24
sqlite: use BEGIN IMMEDIATE by default

crawshaw:
This avoids deadlocks caused by ATTACHed database lock ordering.

Test adapted from one @dsnet wrote.


bradfitz (approved):

dsnet (approved):
Nice :)

PR #35, opened 2022-04-02
sqlite: call initTx from BeginTx for write transactions

maisem:
This makes BeginTx call `BEGIN IMMEDIATE` immediately for write
transactions, and prevents stale txs.

Signed-off-by: Maisem Ali <maisem@tailscale.com>


crawshaw (approved):
FWIW I think we can rip out this optimization.

maisem:
> FWIW I think we can rip out this optimization.

Done, PTAL.

crawshaw (approved):

PR #36, opened 2022-05-15
sqlstats: add tracer that prints executed SQL queries

crawshaw:


sqlstats/sqlstats.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (t *Tracer) Query(prepCtx context.Context, id sqlite.TraceConnID, query string, duration time.Duration, err error) {
+	stats := t.queryStats(query)
+
+	atomic.AddInt64(&stats.count, 1)
+	atomic.AddInt64(&stats.duration, int64(duration))
+	if err != nil {
+		atomic.AddInt64(&stats.errors, 1)
+	}
+}
```
bradfitz, line 83: so this part runs on each query? a sync.Map for `t.queries` is probably a reasonable use of sync.Map considering that once it's seen each unique query it's never written to

crawshaw: Note that only the first time a query is seen is the write mutex held. After that, only the read lock is needed. Given that means ~never holding the write lock in steady state I think this works out more efficient than a sync.Map.

bradfitz: A sync.Map fast path is an atomic load. This PR's fast path is two atomic adds.

A sync.Map would be faster.

bradfitz (approved):

sqlstats/sqlstats.go
crawshaw: Added a TODO, as I'm trying to clean my current open PRs, will come back to it.

PR #40, opened 2022-05-21
sqlite: store nil []byte as NULL

### LINKED ISSUE #39: No distinction between inserting nil vs empty []byte slices
Currently driver stores both nil and empty byte slices as non-NULL empty blob values.

I'm not sure whether it's a deliberate design choice, or just a corner case nobody hit yet.
With other database drivers I got used to the following behavior:

- insert nil `[]byte` as NULL;
- insert non-nil zero-sized `[]byte` as zero-sized blob.

Other Go sqlite driver implementations make such distinction: in both github.com/mattn/go-sqlite3 and modernc.org/sqlite (note that its versions [v1.16.0, v1.17.2] are buggy in this respect) if you put a nil byte slice into a column, you get a NULL value back.

One partucular use case where it may be useful to make such a distinction is when dealing with optional stored json blobs and some constraints, like such:

```sql
CREATE TABLE t(
  Path TEXT PRIMARY KEY NOT NULL,
  Doc TEXT NOT NULL,
  Tags TEXT check(Tags is NULL OR(json_valid(Tags) AND json_type(Tags)='array'))
)
```

Please see this test:

```go
func TestNilAndEmptyByteSlices(t *testing.T) {
	db := sql.OpenDB(sqlite.Connector("file:/dbname?vfs=memdb", nil, nil))
	defer db.Close()
	if _, err := db.Exec(`CREATE TABLE t(v TEXT)`); err != nil {
		t.Fatal(err)
	}
	for _, c := range [...]struct {
		val     []byte
		notNull bool
	}{
		{[]byte{}, true},
		{nil, false},
	} {
		if _, err := db.Exec(`DELETE FROM t`); err != nil {
			t.Fatalf("emptying the table: %v", err)
		}
		if _, err := db.Exec(`INSERT INTO t(v) VALUES(?)`, c.val); err != nil {
			t.Fatal(err)
		}
		var res sql.NullString
		if err := db.QueryRow(`SELECT v FROM t LIMIT 1`).Scan(&res); err != nil {
			t.Fatalf("reading result for original value of %#v: %v", c.val, err)
		}
		if res.Valid != c.notNull {
			t.Fatalf("got unexpected result %#v for original value of %#v", res, c.val)
		}
	}
}
```

artyom:
Closes #39

Signed-off-by: Artyom Pervukhin <github@artyom.dev>

### DIFF
--- a/sqlite.go
+++ b/sqlite.go
@@ -600,6 +600,9 @@ func (s *stmt) bindBasic(debugName interface{}, ordinal int, v interface{}) (fou
 	case float64:
 		return true, s.stmt.BindDouble(ordinal, v)
 	case []byte:
+		if v == nil {
+			return true, s.stmt.BindNull(ordinal)
+		}
 		if len(v) == 0 {
 			return true, s.stmt.BindZeroBlob64(ordinal, 0)
 		} else {
--- a/sqlite_test.go
+++ b/sqlite_test.go
@@ -312,6 +312,34 @@ func TestEmptyString(t *testing.T) {
 	}
 }
 
+func TestNilAndEmptyByteSlices(t *testing.T) {
+	db := openTestDB(t)
+	if _, err := db.Exec(`CREATE TABLE t(v TEXT)`); err != nil {
+		t.Fatal(err)
+	}
+	for _, c := range [...]struct {
+		val     []byte
+		notNull bool
+	}{
+		{[]byte{}, true},
+		{nil, false},
+	} {
+		if _, err := db.Exec(`DELETE FROM t`); err != nil {
+			t.Fatalf("emptying the table: %v", err)
+		}
+		if _, err := db.Exec(`INSERT INTO t(v) VALUES(?)`, c.val); err != nil {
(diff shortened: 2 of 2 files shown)


artyom:
@crawshaw do you think it's a reasonable change?

DentonGentry:
I think the main issue is just lack of cycles: this repository is used in Tailscale's backend systems and really aimed specifically at those internal needs. You're welcome to fork it to do what you need.

crawshaw:
Hi @artyom, the other thing going on here is we're not sure what the future of the driver package is. We're thinking of moving away from database/sql and using the cgosqlite package via another mechanism. So it's hard to find the time to fix something in the driver that will probably require downstream changes in a non-oss repository right now.

Hopefully we'll have this figured out in a couple of weeks.

artyom:
Thank you for the feedback! If this isn't something that contributes to this library use case, let's close this PR and a related issue. Package documentation is very up-front about its “internal use only” nature, so there's nothing to see there indeed. 😅

PR #42, opened 2022-05-31
cgosqlite: cache column name strings forever

bradfitz:
The set of column names in a program is relatively bounded; cache them
in memory forever to save on some per-query allocs.

Benchmark (of 5 row result set) doesn't show huge CPU/time win, but
alloc reduction does show up. Questionable whether this matters, but I
saw it in profiles.

```
name              old time/op    new time/op    delta
ReadStringRows-8    5.62µs ± 0%    5.67µs ± 0%   +0.88%  (p=0.000 n=9+9)

name              old alloc/op   new alloc/op   delta
ReadStringRows-8    1.62kB ± 0%    1.30kB ± 0%  -19.80%  (p=0.000 n=10+10)

name              old allocs/op  new allocs/op  delta
ReadStringRows-8      79.0 ± 0%      69.0 ± 0%  -12.66%  (p=0.000 n=10+10)
```

### DIFF
--- a/cgosqlite/cgosqlite.go
+++ b/cgosqlite/cgosqlite.go
@@ -54,6 +54,7 @@ package cgosqlite
 // #include "cgosqlite.h"
 import "C"
 import (
+	"sync"
 	"time"
 	"unsafe"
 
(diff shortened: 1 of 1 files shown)


bradfitz:
This was wrong. It was supposed to be on ColumnName, not ColumnText. But ColumnName isn't hot (or even used) by us currently so not really worth fixing yet.

PR #48, opened 2022-06-12
sqlitepool: new package, implements a pool

crawshaw:
This is intended as a replacement for our internal sqlitedb package
pool and its TxWrapper.

It needs a little more machinery to be dropped in that's not covered in
this PR, in particular:
- helper functions that mimic ExecContext/QueryContext/QueryRowContext
  to make replacing the existing uses of TxWrappers easy
- a little more Tracer plumbing, so we can get SQL query exec times
  out in the same format we are used to them now
- a few other missing counters

This PR implements the Tx/Rx object split we've discussed a few times.
It will take a bit of work to make good use of that internally. But I
figured it's worth getting in now.

After we've replaced the current pool, I plan to have the code generator for queries work directly on top of this pool.

Any other features you want to see in a pool?

### DIFF
--- a/sqlitepool/sqlitepool.go
+++ b/sqlitepool/sqlitepool.go
@@ -121,6 +121,7 @@ func (p *Pool) Close() error {
 var errPoolClosed = fmt.Errorf("%w: sqlitepool closed", context.Canceled)
 
 // BeginTx creates a writable transaction using BEGIN IMMEDIATE.
+// The parameter why is passed to the Tracer for debugging.
 func (p *Pool) BeginTx(ctx context.Context, why string) (*Tx, error) {
 	select {
 	case <-p.closed:
@@ -134,7 +135,7 @@ func (p *Pool) BeginTx(ctx context.Context, why string) (*Tx, error) {
 			p.tracer.BeginTx(ctx, conn.id, why, false, err)
 		}
 		if err != nil {
-			p.rwConnFree <- conn
+			p.rwConnFree <- conn // can't block, buffer is big enough
 			return nil, err
 		}
 		return tx, nil
@@ -142,6 +143,7 @@ func (p *Pool) BeginTx(ctx context.Context, why string) (*Tx, error) {
 }
 
 // BeginRx creates a read-only transaction.
+// The parameter why is passed to the Tracer for debugging.
 func (p *Pool) BeginRx(ctx context.Context, why string) (*Rx, error) {
 	select {
 	case <-p.closed:
@@ -155,7 +157,7 @@ func (p *Pool) BeginRx(ctx context.Context, why string) (*Rx, error) {
 			p.tracer.BeginTx(ctx, conn.id, why, true, err)
 		}
 		if err != nil {
-			p.roConnsFree <- conn
+			p.roConnsFree <- conn // can't block, buffer is big enough
 			return nil, err
 		}
 		return &Rx{conn: conn}, nil
--- a/sqlitepool/sqlitepool.go
+++ b/sqlitepool/sqlitepool.go
@@ -1,0 +1,318 @@
+// Package sqlitepool implements a pool of SQLite database connections.
+package sqlitepool
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/tailscale/sqlite/cgosqlite"
+	"github.com/tailscale/sqlite/sqliteh"
+)
+
+// A Pool is a fixed-size pool of SQLite database connections.
+// One is reserved for writable transactions, the others are
+// used for read-only transactions.
+type Pool struct {
+	poolSize    int
+	rwConnFree  chan *conn // cap == 1
+	roConnsFree chan *conn // cap == poolSize-1
+	tracer      sqliteh.Tracer
+	closed      chan struct{}
+}
+
+type conn struct {
+	pool  *Pool
+	db    sqliteh.DB
+	stmts map[string]sqliteh.Stmt // persistent statements on db
+	id    sqliteh.TraceConnID
+}
+
+// NewPool creates a Pool of poolSize database connections.
+//
+// For each connection, initFn is called to initialize the connection.
+// Tracer is used to report statistics about the use of the Pool.
+func NewPool(filename string, poolSize int, initFn func(sqliteh.DB) error, tracer sqliteh.Tracer) (p *Pool, err error) {
+	defer func() {
+		if err != nil {
+			err = fmt.Errorf("sqlitepool.NewPool: %w", err)
+			select {
+			case conn := <-p.rwConnFree:
+				conn.db.Close()
+			default:
+			}
+			for conn := range p.roConnsFree {
+				conn.db.Close()
+			}
+		}
+	}()
+	if poolSize < 2 {
+		return nil, fmt.Errorf("poolSize=%d is too small", poolSize)
+	}
+	p = &Pool{
+		poolSize:    poolSize,
+		rwConnFree:  make(chan *conn, 1),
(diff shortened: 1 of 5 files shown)


sqlitepool/sqlitepool.go
```diff
@@ -0,0 +1,288 @@
+// Package sqlitepool implements a Pool of SQLite database connections.
```
bradfitz, line 1: the capitalized Pool reads oddly

sqlitepool/sqlitepool.go
```diff
@@ -0,0 +1,288 @@
...
+
+	"github.com/tailscale/sqlite/cgosqlite"
+	"github.com/tailscale/sqlite/sqliteh"
+)
+
+// A Pool is a fixed-size pool of SQLite database connections.
+// One is reserved for writable transactions, the others are
+// used for read-only transactions.
+type Pool struct {
+	rwConnFree chan *conn // one-element chnnael
```
bradfitz, line 20: typo: channel

sqlitepool/sqlitepool.go
```diff
@@ -0,0 +1,288 @@
...
+//
+// A Tx contains an embedded Rx, which can be used to pass to functions
+// that want to perform read-only queries on the writable Tx.
+type Tx struct {
+	*Rx
+
+	OnCommit func()
+}
+
+// Rollback executes ROLLBACK and cleans up the Tx.
```
bradfitz, line 258: clarify it's a no-op if already Rollback/Commit.

on Commit too.

sqlitepool/sqlitepool.go
```diff
@@ -0,0 +1,288 @@
...
+	}
+	err := rx.rollback()
+	rx.conn.pool.roConnsFree <- rx.conn
+	rx.conn = nil
+	if err != nil {
+		panic(err)
+	}
+}
+
+// Tx is a writable SQLite database transaction.
```
bradfitz, line 248: clarify that nothing's safe for concurrent use.

bradfitz (commented):
Brief skim, but nice. Happy to have a formal Rx.

sqlitepool/sqlitepool.go
```diff
@@ -0,0 +1,288 @@
...
+		return &Rx{conn: conn}, nil
+	}
+}
+
+// Rx is a read-only transaction.
+type Rx struct {
+	conn *conn
+	inTx bool // true if this Rx is embedded in a writable Tx
+
+	OnRollback func()
```
bradfitz, line 153: docs. optional I assume.

crawshaw:
This PR now includes a second commit filled with basic implementations of the Exec/QueryRow/Query, like `database/sql`.

Long-term I'm hoping to replace these with generated access to statements (like sqlc, but without the allocations), but this lets us get sqlitepool into our server with mostly-mechanical perl one-liners.

crawshaw:
Ready for review.

See it in action here: https://github.com/tailscale/corp/pull/5720

sqlitepool/sqlitepool.go
```diff
@@ -0,0 +1,318 @@
...
+	return err
+}
+
+func (p *Pool) Close() error {
+	select {
+	case <-p.closed:
+		return errors.New("pool already closed")
+	default:
+	}
+	close(p.closed)
```
bradfitz, line 106: Note that this pattern doesn't work from concurrent goroutine callers.

sqlitepool/sqlitepool.go
```diff
@@ -0,0 +1,318 @@
...
+		c := <-p.roConnsFree
+		err2 := c.close()
+		if err == nil {
+			err = err2
+		}
+	}
+	return err
+}
+
+var errPoolClosed = fmt.Errorf("%w: sqlitepool closed", context.Canceled)
```
bradfitz, line 121: I suppose this is a canceled error. But why?

sqlitepool/sqlitepool.go
```diff
@@ -0,0 +1,318 @@
...
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	case conn := <-p.rwConnFree:
+		tx := &Tx{Rx: &Rx{conn: conn, inTx: true}}
+		err := tx.Exec("BEGIN IMMEDIATE;")
+		if p.tracer != nil {
+			p.tracer.BeginTx(ctx, conn.id, why, false, err)
+		}
+		if err != nil {
+			p.rwConnFree <- conn
```
bradfitz, line 137: also select on closed here?

Or I guess a return of a conn to its buffered pool can't block.

Maybe a `// can't block` comment to assure future readers

sqlitepool/sqlitepool.go
```diff
@@ -0,0 +1,318 @@
...
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	case conn := <-p.roConnsFree:
+		rx := &Rx{conn: conn}
+		err := rx.Exec("BEGIN;")
+		if p.tracer != nil {
+			p.tracer.BeginTx(ctx, conn.id, why, true, err)
+		}
+		if err != nil {
+			p.roConnsFree <- conn
```
bradfitz, line 158: same

sqlitepool/sqlitepool.go
```diff
@@ -0,0 +1,318 @@
...
+		}
+		if err != nil {
+			p.rwConnFree <- conn
+			return nil, err
+		}
+		return tx, nil
+	}
+}
+
+// BeginRx creates a read-only transaction.
```
bradfitz, line 144: document `why` here and above?

bradfitz (approved):

sqlitepool/sqlitepool.go
```diff
@@ -0,0 +1,318 @@
...
+	}
+	p = &Pool{
+		poolSize:    poolSize,
+		rwConnFree:  make(chan *conn, 1),
+		roConnsFree: make(chan *conn, poolSize-1),
+		tracer:      tracer,
+		closed:      make(chan struct{}),
+	}
+	for i := 0; i < poolSize; i++ {
+		db, err := cgosqlite.Open(filename, sqliteh.OpenFlagsDefault, "")
```
burdiyan, line 61: Seems like there's no way to pass open flags or VFS using NewPool?

sqlitepool/sqlitepool.go
```diff
@@ -0,0 +1,318 @@
...
+		c := &conn{
+			pool:  p,
+			db:    db,
+			stmts: make(map[string]sqliteh.Stmt),
+			id:    sqliteh.TraceConnID(i),
+		}
+		if i == 0 {
+			p.rwConnFree <- c
+		} else {
+			if err := ExecScript(c.db, "PRAGMA query_only=true"); err != nil {
```
burdiyan, line 77: Isn't `SQLITE_OPEN_READONLY` open flag a bit more appropriate for read-only connections? I believe SQLite does some optimizations for read-only connections which may improve performance. But I'm not sure about it.

sqlitepool/sqlitepool.go
crawshaw: true, but I don't expect that here

sqlitepool/sqlitepool.go
crawshaw: so I can `errors.Is` on it for a generic canceled error, but still figure out where it came from when it gets printed out

sqlitepool/sqlitepool.go
crawshaw: not yet. may add in the future.

sqlitepool/sqlitepool.go
crawshaw: good idea, will do in a followup

PR #49, opened 2022-06-12
sqlstats, sqliteh: connection stats, and plumb why

crawshaw:
This uses the sqlstats tracer to track current active transactions, and
how long they have been running for.

The database/sql driver does not have a Tx reason to plumb into 'why',
but the new sqlitepool does.

### DIFF
--- a/sqlstats/sqlstats.go
+++ b/sqlstats/sqlstats.go
@@ -4,6 +4,7 @@ package sqlstats
 import (
 	"context"
 	"fmt"
+	"html"
 	"net/http"
 	"net/url"
 	"sort"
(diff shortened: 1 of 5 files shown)


sqlstats/sqlstats.go
```diff
@@ -90,10 +107,78 @@ func (t *Tracer) Query(prepCtx context.Context, id sqliteh.TraceConnID, query st
...
+	w.WriteHeader(200)
+	fmt.Fprintf(w, "<!DOCTYPE html><html><title>sqlite conns</title><body>\n")
+	fmt.Fprintf(w, "<p>outstanding sqlite transactions: %d</p>\n", len(summary))
+	fmt.Fprintf(w, "<pre>\n")
+	for _, s := range summary {
+		rw := ""
+		if !s.readOnly {
+			rw = " read-write"
+		}
+		fmt.Fprintf(w, "\n\t%s (%v)%s", s.name, now.Sub(s.start).Round(time.Millisecond), rw)
```
bradfitz, line 178: html.EscapeString(s.name)

bradfitz (approved):

PR #51, opened 2022-07-14
sqlite: reuse existing persistence contexts

raggi:
I observed a hot pattern of:

```go
  func parent(ctx context.Context) {
    ctx = sqlite.WithPersist(ctx)
    ...
    child(ctx, ...)
  }

  func child(ctx context.Context) {
    tx.PrepareContext(sqlite.WithPersist(ctx, ...))
    ...
    tx.PrepareContext(sqlite.WithPersist(ctx, ...))
    ...
  }
```

This pattern was producing additional garbage from context creation that
is now avoided with this check.


crawshaw:
FYI: this all goes away in the sqlitepool PRs I have out. Might try to get them in tomorrow.

raggi:
Sounds good :)

I just noticed this in passing.

PR #55, opened 2023-03-02
cgosqlite: add DB.SetWALHook

bradfitz:


bradfitz:
halp. I apparently suck at cgo more than I thought. Can't get it happy.

bradfitz:
okay, I think this works now? PTAL.

cgosqlite/cgosqlite.go
```diff
@@ -150,6 +170,15 @@ func (db *DB) AutoCheckpoint(n int) error {
 	return errCode(res)
 }
 
+func (db *DB) SetWALHook(f func(dbName string, pages int)) {
+	walHookMu.Lock()
+	handle := C.uintptr_t(len(walHooks))
+	walHooks[handle] = f
+	walHookMu.Unlock()
+
+	C.our_sqlite3_wal_hook(db.db, handle)
```
crawshaw, line 179: There are two pieces of cleanup to add at some point:

1. On `SetWALHook(nil)` delete the map entry. (minor)
2. On Close, delete the map entry, to avoid leaking closures

TODOs would be fine.

crawshaw (approved):

cgosqlite/cgosqlite.go
bradfitz: Fixed.

bradfitz:
PTAL. Simplified.

cgosqlite/cgosqlite.h
```diff
@@ -70,3 +74,14 @@ static int reset_and_clear(sqlite3_stmt* stmt, struct timespec* start, int64_t*
...
 }
+
+int walCallbackGo(sqlite3 *db, char *dbName, int pages);
+
+static int wal_callback_gen(void *userData, sqlite3 *db, const char *dbName,
+                            int pages) {
+	return walCallbackGo(db, (char *)dbName, pages);
+}
+
+static void our_sqlite3_wal_hook(sqlite3* db) {
```
maisem, line 85: maybe `ts_sqlite3_wal_hook`

bradfitz: renamed both

bradfitz:
PTAL

cgosqlite/cgosqlite.go
```diff
@@ -150,6 +171,15 @@ func (db *DB) AutoCheckpoint(n int) error {
 	return errCode(res)
 }
 
+func (db *DB) SetWALHook(f func(dbName string, pages int)) {
```
maisem, line 174: i think this needs to be on `*conn`

maisem: oh, i guess i don't fully understand this layer.

maisem (approved):

PR #56, opened 2023-03-04
sqlite: remove some allocs

bradfitz:


sqlite_test.go
```diff
@@ -835,6 +836,54 @@ func TestSetWALHook(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	n := testing.AllocsPerRun(10000, func() {
+		if _, err := stmt.Exec(); err != nil {
+			t.Fatal(err)
+		}
+	})
+	const maxAllocs = 3 // as of Go 1.20
+	if n > maxAllocs {
```
maisem, line 855: What was it before?

maisem (approved):

sqlite_test.go
bradfitz: 5 IIRC?

bradfitz:
@crawshaw, you fine with this bumping minimum required Go version to 1.20?

Feel free to merge if happy.

crawshaw (approved):
looks right!

PR #58, opened 2023-03-08
add QueryBinary, an alloc-free way to read all rows into a buffer

bradfitz:
WIP; goal is alloc-free reads of a query into a Go-provided buffer. Go code can then parse the simple binary format and alloc if needed (doing its own cache lookups, including alloc-free m[string([]byte)] lookups, and returning existing Views if data is unmodified)

(Counterproposal to #43)


cgosqlite/cgosqlite.go
```diff
@@ -401,6 +403,24 @@ func (stmt *Stmt) ColumnDeclType(col int) string {
 	return res
 }
 
+func (stmt *Stmt) StepAllBinary(dstBuf []byte) (n int, err error) {
```
crawshaw, line 406: :ohno:

you could, now you're going this route... make this `AppendAllBinary(dst []byte) ([]byte, int, error)`

and it could loop over the C function, with it returning every time a row doesn't fit for AppendAllBinary to grow dst, then pass it back in to read more rows.

that way you get a clean Go api, an amortized number of C calls over the growing size of dst, and only a single pass through the query

bradfitz: Look at the caller of this. That's where clean Go API will go. This is the lower level wrapper.

crawshaw: I was suggesting that if you move that logic in here, instead of calling ResetAndClear every time you need to grow the array, you just have to start encoding the current row again. That way you only have to step through a query once. (Which is nice if someone throws in a huge complex join that requires sqlite go off and do some huge amount of pre-work.)

bradfitz: I don't want to attempt to bail out of step iteration in C, go back to Go, and then try to get back into C where I left off. Just not worth the complexity.

All or nothing.

Callers can arrange good enough buffer sizes in pools.

PR #65, opened 2023-03-26
cgosqlite: remove allocs in bind_{int64,null,double}

bradfitz:
Updates tailscale/corp#9919


bradfitz:
I found this while adding bind parameters to https://github.com/tailscale/sqlite/pull/58 

BTW, https://github.com/tailscale/sqlite/pull/58 is down to zero allocs now.

maisem (approved):

PR #66, opened 2023-04-09
sqlite: don't cache sqlite column types between rows

### LINKED ISSUE #32: Issues with interactions with custom Scanners
I am working on something involving this database driver that makes use of the sql.Null\* types. When making queries that have a mix of NULL and non-NULL values in a column, only NULL values are returned. Consider this SQLite shell session:

```
sqlite> CREATE TABLE foo ( data TEXT );
sqlite> INSERT INTO foo(data) VALUES ('bar');
sqlite> INSERT INTO foo(data) VALUES (NULL);
```

When I run this in Go and query `SELECT data FROM foo ORDER BY rowid`, I get the following (after scanning it into a slice of sql.NullString):

```go
[]sql.NullString{
  {String: "bar", Valid: true},
  {String: "", Valid: true},
}
```

Which is confusing because the SQLite shell returns the following for the moral equivalent of that datatype:

```
sqlite> SELECT data, data IS NOT NULL AS valid FROM foo ORDER BY rowid;
| data | valid |
|------|-------|
| bar  | 1     |
|      | 0     |
```

Here is a minimal reproduction case:

```go
func TestSQLiteBug(t *testing.T) {
	db := sql.OpenDB(sqlite.Connector(":memory:", func(ctx context.Context, conn driver.ConnPrepareContext) error { return nil }, nil))
	err := db.Ping()
	if err != nil {
		t.Fatal(err)
	}

	_, err = db.Exec("CREATE TABLE foo ( data TEXT )")
	if err != nil {
		t.Fatal(err)
	}

	db.Exec("INSERT INTO foo(data) VALUES ('bar')")
	db.Exec("INSERT INTO foo(data) VALUES (NULL)")

	var result []sql.NullString
	rows, err := db.Query("SELECT data FROM foo ORDER BY rowid")
	if err != nil {
		t.Fatal(err)
	}
	defer rows.Close()

	for rows.Next() {
		var val sql.NullString
		err := rows.Scan(&val)
		if err != nil {
			t.Fatal(err)
		}
		result = append(result, val)
	}

	if len(result) != 2 {
		t.Fatalf("wanted 2 results, not %d", len(result))
	}

	first := result[0]
	second := result[1]

	t.Logf("first:  %#+v", first)
	t.Logf("second: %#+v", second)

	if !first.Valid {
		t.Error("first is not valid but we wanted it to be")
	}

	if first.String != "bar" {
		t.Errorf("first is not \"bar\", got: %q", first.String)
	}

	if second.Valid {
		t.Error("second is valid but we wanted it to not be")
	}
}
```

This fails with the following message:

```
=== RUN   TestSQLiteBug
    sqlite_date_test.go:101: first:  sql.NullString{String:"bar", Valid:true}
    sqlite_date_test.go:102: second: sql.NullString{String:"", Valid:true}
    sqlite_date_test.go:113: second is valid but we wanted it to not be
--- FAIL: TestSQLiteBug (0.00s)
```

There have been other issues when adding left joins into the mix, but details will be provided in the comments when I am able to get a minimal reproduction.

bradfitz:
If we had a NULL in the first row, we were assuming all rows thereafter had a NULL in that column.

Further, SQLite lets you have different types per row even within the same column, so caching the column type was kinda wrong in general anyway.

Presumably this was cached previously for performance reasons, but the handful of commits prior to this one made the cgo calls alloc-free, so this is much cheaper than it used to be.

                             │   before    │                after                │
                             │   sec/op    │   sec/op     vs base                │
    WALHookAndExec-8           2.316µ ± 2%   2.327µ ± 1%        ~ (p=0.254 n=10)
    Persist-8                  1.980µ ± 1%   2.037µ ± 1%   +2.85% (p=0.001 n=10)
    QueryRows100MixedTypes-8   36.44µ ± 2%   46.14µ ± 1%  +26.64% (p=0.000 n=10)
    EmptyExec-8                357.0n ± 2%   356.6n ± 1%        ~ (p=0.565 n=10)
    BeginTxNoop-8              9.178µ ± 0%   9.204µ ± 0%        ~ (p=0.239 n=10)
    geomean                    3.529µ        3.726µ        +5.56%

                             │    before    │                 after                 │
                             │     B/op     │     B/op      vs base                 │
    WALHookAndExec-8             48.00 ± 0%     48.00 ± 0%       ~ (p=1.000 n=10) ¹
    Persist-8                    456.0 ± 0%     424.0 ± 0%  -7.02% (p=0.000 n=10)
    QueryRows100MixedTypes-8   2.719Ki ± 0%   2.688Ki ± 0%  -1.15% (p=0.000 n=10)
    EmptyExec-8                  32.00 ± 0%     32.00 ± 0%       ~ (p=1.000 n=10) ¹
    BeginTxNoop-8                624.0 ± 0%     624.0 ± 0%       ~ (p=1.000 n=10)
    geomean                      261.2          256.9       -1.67%
    ¹ all samples are equal

                             │   before   │                after                │
                             │ allocs/op  │ allocs/op   vs base                 │
    WALHookAndExec-8           2.000 ± 0%   2.000 ± 0%       ~ (p=1.000 n=10) ¹
    Persist-8                  11.00 ± 0%   11.00 ± 0%       ~ (p=1.000 n=10) ¹
    QueryRows100MixedTypes-8   107.0 ± 0%   107.0 ± 0%       ~ (p=1.000 n=10) ¹
    EmptyExec-8                1.000 ± 0%   1.000 ± 0%       ~ (p=1.000 n=10) ¹
    BeginTxNoop-8              13.00 ± 0%   13.00 ± 0%       ~ (p=1.000 n=10) ¹
    geomean                    7.891        7.891       +0.00%
    ¹ all samples are equal

Fixes #32

### DIFF
--- a/sqlite.go
+++ b/sqlite.go
@@ -374,7 +374,6 @@ type stmt struct {
 	prepCtx context.Context // the context provided to prepare, for tracing
 
 	// filled on first step only if persist==true
-	colTypes     []sqliteh.ColumnType
 	colDeclTypes []colDeclType
 	colNames     []string
 }
@@ -634,7 +633,6 @@ type rows struct {
 	colNames []string // filled on call to Columns
 
 	// Filled on first call to Next.
-	colTypes     []sqliteh.ColumnType
 	colDeclTypes []colDeclType
 }
 
@@ -681,28 +679,24 @@ func (r *rows) Next(dest []driver.Value) error {
 		return io.EOF
 	}
 
-	if r.colTypes == nil {
-		if r.stmt.colTypes != nil {
-			r.colTypes = r.stmt.colTypes
-			r.colDeclTypes = r.stmt.colDeclTypes
-		} else {
-			colCount := r.stmt.stmt.ColumnCount()
-			r.colTypes = make([]sqliteh.ColumnType, colCount)
-			r.colDeclTypes = make([]colDeclType, colCount)
-			for i := range r.colTypes {
-				r.colTypes[i] = r.stmt.stmt.ColumnType(i)
-				r.colDeclTypes[i] = colDeclTypeFromString(r.stmt.stmt.ColumnDeclType(i))
-			}
-			if r.stmt.persist {
-				r.stmt.colTypes = r.colTypes
-				r.stmt.colDeclTypes = r.colDeclTypes
-			}
+	if r.colDeclTypes == nil {
+		r.colDeclTypes = r.stmt.colDeclTypes
+	}
+	if r.colDeclTypes == nil {
+		colCount := r.stmt.stmt.ColumnCount()
+		r.colDeclTypes = make([]colDeclType, colCount)
+		for i := range r.colDeclTypes {
+			r.colDeclTypes[i] = colDeclTypeFromString(r.stmt.stmt.ColumnDeclType(i))
+		}
+		if r.stmt.persist {
+			r.stmt.colDeclTypes = r.colDeclTypes
 		}
 	}
 
 	for i := range dest {
+		colType := r.stmt.stmt.ColumnType(i)
 		if r.colDeclTypes[i] == declTypeDateOrTime {
-			switch r.colTypes[i] {
+			switch colType {
 			case sqliteh.SQLITE_INTEGER:
 				v := r.stmt.stmt.ColumnInt64(i)
 				dest[i] = time.Unix(v, 0)
@@ -729,7 +723,7 @@ func (r *rows) Next(dest []driver.Value) error {
 			}
 			continue
 		}
-		switch r.colTypes[i] {
+		switch colType {
 		case sqliteh.SQLITE_INTEGER:
 			val := r.stmt.stmt.ColumnInt64(i)
 			if r.colDeclTypes[i] == declTypeBoolean {
--- a/sqlite_test.go
+++ b/sqlite_test.go
@@ -857,6 +857,88 @@ func TestSetWALHook(t *testing.T) {
 	}
 }
 
+// Tests that we don't remember the SQLite column types of the first row of the
+// result set (notably the "NULL" type) and re-use it for all subsequent rows
+// like we used to.
+func TestNoStickyColumnTypes(t *testing.T) {
+	db := openTestDB(t)
+	exec(t, db, "CREATE TABLE t (id INTEGER PRIMARY KEY, v1 ANY, v2 ANY)")
+
+	type row []any
+	r := func(v ...any) row { return v }
+	rs := func(v ...row) []row { return v }
+	tests := []struct {
+		name string
+		rows []row
+	}{
+		{"no-null", rs(
+			r("a", "b"),
+			r("foo", "bar"))},
+		{"only-null", rs(
+			r(nil, nil))},
+		{"null-after-string", rs(
+			r("a", "b"),
+			r(nil, "bar"))},
+		{"string-after-null", rs(
+			r(nil, "b"),
+			r("foo", "bar"))},
+		{"null-after-int", rs(
+			r(101, 102),
+			r(nil, 202))},
+		{"int-after-null", rs(
+			r(nil, 102),
+			r(201, 202))},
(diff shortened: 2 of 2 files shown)


creachadair (approved):
🤯 TIL that SQLite allows the type to vary across rows. Is that only within affinities, or would it (say) allow TEXT and INTEGER to mix?

raggi:
> 🤯 TIL that SQLite allows the type to vary across rows. Is that only within affinities, or would it (say) allow TEXT and INTEGER to mix?

Switching from default to SQLite strict tables is a bit like switching to innodb in MySQL. https://www.sqlite.org/stricttables.html

Nullability would still be a thing though.

bradfitz:
> 🤯 TIL that SQLite allows the type to vary across rows. Is that only within affinities, or would it (say) allow TEXT and INTEGER to mix?

Anything. Look at the `changing-types-within-a-column-between-rows` test case that I'd added which shows changing between string, integer, and null.

creachadair:
> > 🤯 TIL that SQLite allows the type to vary across rows. Is that only within affinities, or would it (say) allow TEXT and INTEGER to mix?
> 
> Anything. Look at the `changing-types-within-a-column-between-rows` test case that I'd added which shows changing between string, integer, and null.

INTEGER/TEXT isn't surprising—the docs promise it will apply standard-ish coercions, which empirically includes parsing integer-shaped strings and rendering integers into strings. The fact that it will also let you jam floats (say) in there was the surprising part to me 😭

bradfitz:
BTW, the benchmark numbers here may look bad but I have a follow-up change which does all the column_type lookups in the same cgo call as the sqlite3_step call. That actually makes it ~faster (barely), comparing from before this commit to after the two commits:

```
                         │   before    │               after2               │
                         │   sec/op    │   sec/op     vs base               │
WALHookAndExec-8           2.316µ ± 2%   2.297µ ± 1%       ~ (p=0.109 n=10)
Persist-8                  1.980µ ± 1%   1.956µ ± 0%  -1.26% (p=0.000 n=10)
QueryRows100MixedTypes-8   36.44µ ± 2%   36.47µ ± 1%       ~ (p=0.289 n=10)
EmptyExec-8                357.0n ± 2%   355.1n ± 1%       ~ (p=0.697 n=10)
BeginTxNoop-8              9.178µ ± 0%   9.178µ ± 2%       ~ (p=0.684 n=10)
geomean                    3.529µ        3.512µ       -0.50%

                         │    before    │                after2                 │
                         │     B/op     │     B/op      vs base                 │
WALHookAndExec-8             48.00 ± 0%     48.00 ± 0%       ~ (p=1.000 n=10) ¹
Persist-8                    456.0 ± 0%     440.0 ± 0%  -3.51% (p=0.000 n=10)
QueryRows100MixedTypes-8   2.719Ki ± 0%   2.703Ki ± 0%  -0.57% (p=0.000 n=10)
EmptyExec-8                  32.00 ± 0%     32.00 ± 0%       ~ (p=1.000 n=10) ¹
BeginTxNoop-8                624.0 ± 0%     624.0 ± 0%       ~ (p=1.000 n=10)
geomean                      261.2          259.1       -0.83%
¹ all samples are equal

                         │   before   │               after2                │
                         │ allocs/op  │ allocs/op   vs base                 │
WALHookAndExec-8           2.000 ± 0%   2.000 ± 0%       ~ (p=1.000 n=10) ¹
Persist-8                  11.00 ± 0%   11.00 ± 0%       ~ (p=1.000 n=10) ¹
QueryRows100MixedTypes-8   107.0 ± 0%   107.0 ± 0%       ~ (p=1.000 n=10) ¹
EmptyExec-8                1.000 ± 0%   1.000 ± 0%       ~ (p=1.000 n=10) ¹
BeginTxNoop-8              13.00 ± 0%   13.00 ± 0%       ~ (p=1.000 n=10) ¹
geomean                    7.891        7.891       +0.00%
¹ all samples are equal
```

crawshaw (approved):

PR #67, opened 2023-04-10
cgosqlite: get column types during Step

bradfitz:
This recovers the performance loss from the prior commit's bug
fix (#66). Compared to before that, we're now even slightly faster at:

```
                             │   before    │               after2               │
                             │   sec/op    │   sec/op     vs base               │
    WALHookAndExec-8           2.316µ ± 2%   2.297µ ± 1%       ~ (p=0.109 n=10)
    Persist-8                  1.980µ ± 1%   1.956µ ± 0%  -1.26% (p=0.000 n=10)
    QueryRows100MixedTypes-8   36.44µ ± 2%   36.47µ ± 1%       ~ (p=0.289 n=10)
    EmptyExec-8                357.0n ± 2%   355.1n ± 1%       ~ (p=0.697 n=10)
    BeginTxNoop-8              9.178µ ± 0%   9.178µ ± 2%       ~ (p=0.684 n=10)
    geomean                    3.529µ        3.512µ       -0.50%

                             │    before    │                after2                 │
                             │     B/op     │     B/op      vs base                 │
    WALHookAndExec-8             48.00 ± 0%     48.00 ± 0%       ~ (p=1.000 n=10) ¹
    Persist-8                    456.0 ± 0%     440.0 ± 0%  -3.51% (p=0.000 n=10)
    QueryRows100MixedTypes-8   2.719Ki ± 0%   2.703Ki ± 0%  -0.57% (p=0.000 n=10)
    EmptyExec-8                  32.00 ± 0%     32.00 ± 0%       ~ (p=1.000 n=10) ¹
    BeginTxNoop-8                624.0 ± 0%     624.0 ± 0%       ~ (p=1.000 n=10)
    geomean                      261.2          259.1       -0.83%
    ¹ all samples are equal

                             │   before   │               after2                │
                             │ allocs/op  │ allocs/op   vs base                 │
    WALHookAndExec-8           2.000 ± 0%   2.000 ± 0%       ~ (p=1.000 n=10) ¹
    Persist-8                  11.00 ± 0%   11.00 ± 0%       ~ (p=1.000 n=10) ¹
    QueryRows100MixedTypes-8   107.0 ± 0%   107.0 ± 0%       ~ (p=1.000 n=10) ¹
    EmptyExec-8                1.000 ± 0%   1.000 ± 0%       ~ (p=1.000 n=10) ¹
    BeginTxNoop-8              13.00 ± 0%   13.00 ± 0%       ~ (p=1.000 n=10) ¹
    geomean                    7.891        7.891       +0.00%
    ¹ all samples are equal
```

### DIFF
--- a/cgosqlite/cgosqlite.go
+++ b/cgosqlite/cgosqlite.go
@@ -272,8 +272,12 @@ func (stmt *Stmt) ColumnTableName(col int) string {
 	return C.GoString((*C.char)(unsafe.Pointer(C.sqlite3_column_table_name(stmt.stmt.ptr(), C.int(col)))))
 }
 
-func (stmt *Stmt) Step() (row bool, err error) {
-	res := C.ts_sqlite3_step(stmt.stmt.int())
+func (stmt *Stmt) Step(colType []sqliteh.ColumnType) (row bool, err error) {
+	var ptr *C.char
+	if len(colType) > 0 {
+		ptr = (*C.char)(unsafe.Pointer(&colType[0]))
+	}
+	res := C.ts_sqlite3_step(stmt.stmt.int(), ptr, C.int(len(colType)))
 	switch res {
 	case C.SQLITE_ROW:
 		return true, nil
--- a/sqlite.go
+++ b/sqlite.go
@@ -630,6 +630,13 @@ type rows struct {
 	stmt   *stmt
 	closed bool
 
+	// colType is the column types for Step to fill on each row. We only use 23
+	// as it packs well with the closed bool byte above (24 bytes total, same as
+	// a slice) and it's uncommon for queries to select so many columns. But if
+	// they do, we still work: we just query the column type via cgo on each
+	// row. So a bit slower, but fine.
+	colType [23]sqliteh.ColumnType
+
 	colNames []string // filled on call to Columns
 
 	// Filled on first call to Next.
@@ -671,7 +678,7 @@ func (r *rows) Next(dest []driver.Value) error {
 	if r.closed {
 		return errors.New("sqlite rows result already closed")
 	}
-	hasRow, err := r.stmt.stmt.Step()
+	hasRow, err := r.stmt.stmt.Step(r.colType[:])
 	if err != nil {
 		return r.stmt.reserr("Rows.Next", err)
 	}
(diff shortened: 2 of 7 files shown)


cgosqlite/cgosqlite.go
```diff
@@ -272,8 +272,12 @@ func (stmt *Stmt) ColumnTableName(col int) string {
 	return C.GoString((*C.char)(unsafe.Pointer(C.sqlite3_column_table_name(stmt.stmt.ptr(), C.int(col)))))
 }
 
-func (stmt *Stmt) Step() (row bool, err error) {
-	res := C.ts_sqlite3_step(stmt.stmt.int())
+func (stmt *Stmt) Step(colType []sqliteh.ColumnType) (row bool, err error) {
+	var ptr *C.char
+	if len(colType) > 0 {
```
creachadair, line 277: Could be `!= 0` (saves an instruction on x86!)

sqlite.go
```diff
@@ -630,6 +630,7 @@ type rows struct {
 	stmt   *stmt
 	closed bool
 
+	colType  [16]sqliteh.ColumnType
```
creachadair, line 633: It would be worth documenting why 16 here too.

creachadair (approved):

cgosqlite/cgosqlite.go
bradfitz: That's not true after the Go compiler's SSA phase and all its rando rules.

Like: IIRC it already has a rule that `len` always returns something non-negative (in the portable SSA rules). And then an x86-specific SSA rule about the thing you just said.

So it applies the two and gets the same result.

You can verify at https://godbolt.org/ comparing,

```
func foo(b []byte) {
    if len(b) > 0 {
        panic("hi")
    }
}
```

vs

```
func foo(b []byte) {
    if len(b) != 0 {
        panic("hi")
    }
}
```

same output on that `len` line:

```
        TESTQ   BX, BX
        JNE     foo_pc40
```

bradfitz: Hah, and if you write `len(b) > -1` that deadcodes away and doesn't generate any code at all, as it compiles to `if true {` effectively.

creachadair: Interesting. I did it using `go tool compile`, and got a compare and a branch vs. a single CBZ, it's quite possible the actual codegen would rewrite the pair. (Stipulated, that even without checking it this isn't really a concern)

sqlite.go
bradfitz: Done. Documented. And changed it to 23 in the process of documenting, because control has one particular hot query with exactly 16 columns. Too close for comfort.

cgosqlite/cgosqlite.go
bradfitz: I also just did it with `go tool compile` and see the same as godbolt.

Were you using a slice? Maybe you weren't hitting the `len(S) is always non-negative` SSA rule?

creachadair: I don't have the code anymore but I think I just used an int, which would explain why it couldn't apply that rule (careless of me).

I promise my goal wasn't to nerdsnipe you 😂

crawshaw (approved):

crawshaw (approved):

PR #71, opened 2023-05-24
sqlite: add metric to track when the driver is called after close

andrew-d:
This is theoretically prohibited by the database/sql package, but after the sql.RawBytes issues we ran into we'd like to verify that there's no other lingering behaviour that's causing us to call SQLite functions after a Close.


sqlite.go
```diff
@@ -172,23 +181,36 @@ type conn struct {
...
+	}
+
 	for q, s := range c.stmts {
 		s.stmt.Finalize()
 		delete(c.stmts, q)
 	}
-	return reserr(c.db, "Conn.Close", "", c.db.Close())
+	err := reserr(c.db, "Conn.Close", "", c.db.Close())
+	c.closed.Store(true)
+	return err
```
bradfitz, line 201: there's a race here if two concurrent Close.

do a CompareAndSwap?

sqlite.go
```diff
@@ -172,23 +181,36 @@ type conn struct {
...
 }
 func (c *conn) PrepareContext(ctx context.Context, query string) (driver.Stmt, error) {
 	persist := ctx.Value(persistQuery{}) != nil
 	return c.prepare(ctx, query, persist)
 }
 
 func (c *conn) prepare(ctx context.Context, query string, persist bool) (s *stmt, err error) {
+	if c.closed.Load() {
+		UsesAfterClose.Add(1)
```
bradfitz, line 210: should we make it a https://pkg.go.dev/expvar#Map and `.Add("prepare", 1)` ?

sqlite.go
```diff
@@ -446,6 +491,10 @@ func (res *stmtResult) LastInsertId() (int64, error) { return res.lastInsertID,
 func (res *stmtResult) RowsAffected() (int64, error) { return res.rowsAffected, nil }
 
 func (s *stmt) QueryContext(ctx context.Context, args []driver.NamedValue) (driver.Rows, error) {
```
bradfitz, line 493: you also want all these `*stmt` methods to check whether the stmt itself was closed (via stmt.Close) before touching the sqliteh.Stmt

sqlite.go
andrew-d: Yeah, I like that; let me update

sqlite.go
andrew-d: How does that work with statements with `s.persist == true`?

andrew-d: After a bunch of testing, I think the right thing to do is to only check if the `*stmt` is closed if `s.persist==false`; just amended to do that. PTAL?

sqlite.go
```diff
@@ -172,23 +181,35 @@ type conn struct {
...
 	readOnly bool
+	closed   atomic.Bool
 }
 
 func (c *conn) Prepare(query string) (driver.Stmt, error) { panic("deprecated, unused") }
 func (c *conn) Begin() (driver.Tx, error)                 { panic("deprecated, unused") }
 func (c *conn) Close() error {
+	// Don't double-close
+	if !c.closed.CompareAndSwap(false, true) {
+		return nil
```
bradfitz, line 192: counter for funsies?

andrew-d:
@bradfitz - just changed the `stmt` logic to set `closed = true` in the `stmt.Close` call and then set it back to `false` when handing out a cached persistent query; PTAL?

sqlite.go
```diff
@@ -172,26 +182,41 @@ type conn struct {
...
 func (c *conn) prepare(ctx context.Context, query string, persist bool) (s *stmt, err error) {
+	if c.closed.Load() {
+		UsesAfterClose.Add("prepare", 1)
+		return nil, ErrClosed
+	}
+
 	query = strings.TrimSpace(query)
 	if s := c.stmts[query]; s != nil {
 		s.prepCtx = ctx
+		s.closed.Store(false)
```
bradfitz, line 219: should this be a CAS for paranoia? Going from false to false would indicate a bug?

sqlite_test.go
```diff
@@ -56,13 +57,31 @@ func configDB(t testing.TB, db *sql.DB) {
 	t.Cleanup(func() { db.Close() })
 }
 
+func getUsesAfterClose() (ret int64) {
```
bradfitz, line 60: cute

sqlite.go
bradfitz: actually, I don't see where this is s.closed.Store(true) on return to the stmts pool?

andrew-d: We do it on `Close` here:
https://github.com/tailscale/sqlite/pull/71/files#diff-3101fd8b5ebefc9739f7efdafa4dd6f04c19db98fc4428623ee619a2773c8501R446

sqlite.go
```diff
@@ -396,6 +447,13 @@ func (s *stmt) NumInput() int {
 }
 
 func (s *stmt) Close() error {
+	// Always set the 'closed' boolean, even for a persisted query; this is
+	// set from false -> true in prepare(), above.
+	if !s.closed.CompareAndSwap(false, true) {
```
bradfitz, line 452: check whether the Stmt's Conn is closed first before doing this CAS to true?

bradfitz (approved):

PR #72, opened 2023-05-26
sqlite: implement driver.SessionResetter and driver.Validator

### LINKED ISSUE #70: Implement driver.SessionResetter and driver.Validator so that conns are retained in database/sql pools
We are missing implementations for [driver.Validator](https://pkg.go.dev/database/sql/driver@go1.20.4#Validator) and [driver.SessionResetter](https://pkg.go.dev/database/sql/driver@go1.20.4#SessionResetter) that are required in order for connections that have errored to be returned to database/sql connection pools.

raggi:
These implementations enable us to maintain connections in database/sql pools.

Fixes #70

### DIFF
--- a/sqlite.go
+++ b/sqlite.go
@@ -275,6 +275,32 @@ func (c *conn) BeginTx(ctx context.Context, opts driver.TxOptions) (driver.Tx, e
 // Raw is so ConnInitFunc can cast to SQLConn.
 func (c *conn) Raw(fn func(any) error) error { return fn(c) }
 
+// ResetSession is called prior to executing a query on the connection
+// if the connection has been used before. If the driver returns ErrBadConn
+// the connection is discarded.
+func (c *conn) ResetSession(ctx context.Context) error {
+	if c.txState != txStateNone {
+		return errors.New("cannot ResetSession on a connection with an active transaction")
+	}
+
+	// ensure that all prepared statements are reset and unbound
+	if c.stmts != nil {
+		for _, s := range c.stmts {
+			if err := s.resetAndClear(); err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+// IsValid is called prior to placing the connection into the
+// connection pool. The connection will be discarded if false is returned.
+func (c *conn) IsValid() bool {
+	return true
+}
+
 type readOnlyKey struct{}
 
 // ReadOnly applies the query_only pragma to the connection.
--- a/sqlite_test.go
+++ b/sqlite_test.go
@@ -20,6 +20,13 @@ import (
 	"github.com/tailscale/sqlite/sqliteh"
 )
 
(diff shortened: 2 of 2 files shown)


sqlite.go
```diff
@@ -275,6 +275,32 @@ func (c *conn) BeginTx(ctx context.Context, opts driver.TxOptions) (driver.Tx, e
...
+func (c *conn) ResetSession(ctx context.Context) error {
+	if c.txState != txStateNone {
+		return errors.New("cannot ResetSession on a connection with an active transaction")
+	}
+
+	// ensure that all prepared statements are reset and unbound
+	if c.stmts != nil {
+		for _, s := range c.stmts {
+			if err := s.resetAndClear(); err != nil {
+				return err
```
maisem, line 290: Should this return ErrBadConn?

raggi: In the above case we could and probably should. In this particular case, that's unclear because:

```
ErrBadConn should be returned by a driver to signal to the sql package that a driver.Conn is in a bad state (such as the server having earlier closed the connection) and the sql package should retry on a new connection.

To prevent duplicate operations, ErrBadConn should NOT be returned if there's a possibility that the database server might have performed the operation. Even if the server sends back an error, you shouldn't return ErrBadConn.

Errors will be checked using errors.Is. An error may wrap ErrBadConn or implement the Is(error) bool method.
```

If we have potentially completed a tx we can not disambiguate between these conditions so we should not match BadConn.

sqlite.go
```diff
@@ -275,6 +275,32 @@ func (c *conn) BeginTx(ctx context.Context, opts driver.TxOptions) (driver.Tx, e
...
+// ResetSession is called prior to executing a query on the connection
+// if the connection has been used before. If the driver returns ErrBadConn
+// the connection is discarded.
+func (c *conn) ResetSession(ctx context.Context) error {
+	if c.txState != txStateNone {
+		return errors.New("cannot ResetSession on a connection with an active transaction")
+	}
+
+	// ensure that all prepared statements are reset and unbound
+	if c.stmts != nil {
```
crawshaw, line 287: nil check is redundant because a range over a nil map does nothing

sqlite.go
```diff
@@ -275,6 +275,32 @@ func (c *conn) BeginTx(ctx context.Context, opts driver.TxOptions) (driver.Tx, e
...
+		for _, s := range c.stmts {
```
crawshaw, line 288: this is almost right. in practice it would basically work. but the problem are statements created with `persist=false`, which don't get put into c.stmts.

crawshaw (commented):
can we generate the sort of error that gets database/sql to call ResetSession? if so, can you write a unit test that exercises this code?

raggi:
I can reverse it. It's absent from the docs. We do do have stats showing that we're not retaining cons. Either way yeah, I'll find a way to make some tests.

Issue #73, opened 2023-06-07
Persistent queries can be returned to two callers


andrew-d:
This code doesn't check whether we've already returned the `*stmt` to a caller, so it's possible that calling `conn.PrepareContext` twice in a row can return the same `*stmt` to two different callers:
https://github.com/tailscale/sqlite/blob/main/sqlite.go#L193-L196

This means that e.g. running a query from each of the two not-actually-different stmts, getting `rows1` and `rows2`, and then calling `rows1.Next` will also step `rows2`.

PR #74, opened 2023-06-07
sqlite: don't return persistent queries to two callers

andrew-d:
This test case used to fail with:

    === RUN   TestPrepareReuse
        sqlite_test.go:1129: rows2: num=2, want 1
    --- FAIL: TestPrepareReuse (0.00s)

This indicates that we were erroneously returning the same "persisted" query twice, which resulted in the two sets of Rows returned affecting each other. Instead, only persist queries in c.stmts on Close, after they've been reset and are ready for re-use.

Updates #73

### DIFF
--- a/sqlite.go
+++ b/sqlite.go
@@ -192,6 +192,9 @@ func (c *conn) prepare(ctx context.Context, query string, persist bool) (s *stmt
 	query = strings.TrimSpace(query)
 	if s := c.stmts[query]; s != nil {
 		s.prepCtx = ctx
+
+		// don't hand the same statement out twice; this is re-added on s.Close
+		delete(c.stmts, query)
 		return s, nil
 	}
 	if c.tracer != nil {
@@ -234,10 +237,12 @@ func (c *conn) prepare(ctx context.Context, query string, persist bool) (s *stmt
 		return s, nil
 	}
 
+	// NOTE: don't add the statement to c.stmts here, since we could return
+	// it to another caller before Close is called; it's added to the
+	// c.stmts map on Close.
 	if c.stmts == nil {
 		c.stmts = make(map[string]*stmt)
 	}
-	c.stmts[query] = s
 	return s, nil
 }
 
@@ -391,8 +396,21 @@ func (s *stmt) NumInput() int {
 }
 
 func (s *stmt) Close() error {
-	if s.persist {
-		return s.reserr("Stmt.Close", s.resetAndClear())
+	// We return this statement to the conn only if it's persistent, and
+	// only if there's not already a statement with the same query already
+	// cached there.
+	shouldPersist := s.persist
+	if shouldPersist {
+		if _, alreadyPersisted := s.conn.stmts[s.query]; alreadyPersisted {
+			shouldPersist = false
+		}
+	}
+	if shouldPersist {
+		err := s.reserr("Stmt.Close", s.resetAndClear())
+		if err == nil {
+			s.conn.stmts[s.query] = s
+		}
+		return err
 	}
 	return s.reserr("Stmt.Close", s.stmt.Finalize())
 }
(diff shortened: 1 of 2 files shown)


sqlite.go
```diff
@@ -391,8 +396,19 @@ func (s *stmt) NumInput() int {
...
 
 func (s *stmt) Close() error {
-	if s.persist {
-		return s.reserr("Stmt.Close", s.resetAndClear())
+	// We return this statement to the conn only if it's persistent, and
+	// only if there's not already a statement with the same query already
+	// cached there.
+	shouldPersist := s.persist
+	if shouldPersist {
+		shouldPersist = s.conn.stmts[s.query] == nil
```
maisem, line 404: How about
```
if _, alreadyPersisted := s.conn.stmts[query]; alreadyPersisted {
  shouldPersist = false
}
```

andrew-d: Sure; done

sqlite_test.go
```diff
@@ -1072,3 +1072,77 @@ func BenchmarkBeginTxNoop(b *testing.B) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	ctx = WithPersist(ctx)
+
+	// Calling PrepareContext twice in a row used to return the same
+	// statement to both callers.
+	const query = "SELECT c FROM t;"
+	stmt1, err := sqlConn.PrepareContext(ctx, query)
```
maisem, line 1104: Try closing these? Deadlocking sounds bad.

maisem (approved):

sqlite.go
```diff
@@ -391,8 +396,19 @@ func (s *stmt) NumInput() int {
 }
 
 func (s *stmt) Close() error {
-	if s.persist {
-		return s.reserr("Stmt.Close", s.resetAndClear())
+	// We return this statement to the conn only if it's persistent, and
+	// only if there's not already a statement with the same query already
+	// cached there.
+	shouldPersist := s.persist
+	if _, alreadyPersisted := s.conn.stmts[s.query]; alreadyPersisted {
```
bradfitz, line 403: no need to check this map if shouldPersist is true.

that penalizes conns that have a mix of persist and non-persist queries (barely)

sqlite_test.go
```diff
@@ -1072,3 +1072,79 @@ func BenchmarkBeginTxNoop(b *testing.B) {
 // TODO(crawshaw): test TextMarshaler
 // TODO(crawshaw): test named types
 // TODO(crawshaw): check coverage
+
+// This tests for https://github.com/tailscale/sqlite/issues/73
```
bradfitz, line 1076: Brief summary and then the URL?

sqlite_test.go
```diff
@@ -1072,3 +1072,79 @@ func BenchmarkBeginTxNoop(b *testing.B) {
...
+		COMMIT;`)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	ctx = WithPersist(ctx)
+
+	// Calling PrepareContext twice in a row used to return the same
+	// statement to both callers.
+	const query = "SELECT c FROM t;"
```
bradfitz, line 1101: FROM t ORDER BY c

otherwise it's not defined the order they're returned, which your tests depends on below

sqlite_test.go
```diff
@@ -1072,3 +1072,79 @@ func BenchmarkBeginTxNoop(b *testing.B) {
...
+	}
+	defer rows1.Close()
+	rows2, err := stmt2.QueryContext(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer rows2.Close()
+
+	assertResult := func(prefix string, rows *sql.Rows, want int) {
+		var num int
```
bradfitz, line 1125: add `t.Helper()` here and then you don't need the `prefix string`?

bradfitz (approved):

PR #76, opened 2023-08-03
cgosqlite: move WAL callback to its own file

### LINKED ISSUE #75: Using dlv fails with compile errors related to _GoStringLen and _GoStringPtr
When I try to use `dlv` in `cgosqlite` or anything that depends on it, I get compiler errors:
```
$ dlv test ./
# github.com/tailscale/sqlite.test
${GOROOT}/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
/usr/bin/ld: /tmp/go-link-2730833985/000000.o: in function `bind_parameter_index':
${HOME}/tailscale/sqlite/cgosqlite/cgosqlite.h:42: undefined reference to `_GoStringLen'
/usr/bin/ld: ${HOME}/tailscale/sqlite/cgosqlite/cgosqlite.h:43: undefined reference to `_GoStringPtr'
collect2: error: ld returned 1 exit status

exit status 1
```

For some reason `go test ./` still succeeds and I haven't figured out what exactly is different.

Anyway looking into this error I eventually came across golang/go#48824 and the very last comment says:
> As far as I can tell `_GoStringLen` and `_GoStringPtr` are declared in the preamble--but only if you don't use `//export`. The `static inline` trick can often work but there is no intention of officially supporting it.

So it looks like the export in `cgosqlite/cgosqlite.go` needs to be moved to a new file.

sailorfrag:
As per the last comment of golang/go#48824, `_GoStringLen` and `_GoStringPtr` are declared in the preamble unless `//export` is used. To fix the problem, the exported WAL callback function needs to be moved to a separate file.

This removes compiler errors under certain conditions. The most easy way to reproduce this is by attempting to use `dlv`. For some reason, a normal `go` build still succeeds but a `dlv` build does not.

Fixes #75

### DIFF
--- a/cgosqlite/cgosqlite.go
+++ b/cgosqlite/cgosqlite.go
@@ -61,24 +61,6 @@ import (
 	"github.com/tailscale/sqlite/sqliteh"
 )
 
-type walHookCb func(dbName string, pages int)
-
-var walHookFunc sync.Map // from *C.sqlite3 to walHookCb
-
-//export walCallbackGo
-func walCallbackGo(db *C.sqlite3, dbNameC *C.char, dbNameLen C.int, pages C.int) C.int {
-	v, _ := walHookFunc.Load(db)
-	hook, _ := v.(walHookCb)
-	if hook == nil {
-		return C.int(0)
-	}
-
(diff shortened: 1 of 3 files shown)


bradfitz (approved):
Nice find!

PR #78, opened 2023-08-20
sqlitepool: remove overly clever panic on bind error

crawshaw:
Turns out the described conditions only hold for binding on standard SQL tables. If you use something like FTS5, you can get SQLITE_MISUSE on bind based on user-provided strings. E.g. if you try to search text for something like "autogroup:shared" it turns out FTS5 parses the ':' on bind and treats the prefix as an SQL table column name and returns a MISUSE error, which you don't want to turn into a panic.

So do the boring thing and return an error.

Updates #cleanup


bradfitz (approved):
Except we still don't use sqlitepool, right?

crawshaw:
In control, no. I was working on a fun app on the weekend using sqlitepool and ran into this.

PR #79, opened 2023-08-20
cgosqlite: import SQLite amalgamation v3.42.0

crawshaw:
Notably, this includes JSON5 support:
https://sqlite.org/json1.html#json5

Updates #cleanup


bradfitz (approved):
I read the changelog from 3.39.4 (https://www.sqlite.org/changes.html). Lot of stuff, but nothing too interesting. Closest interesting thing after JSON5 is:

> The NOT NULL and IS NULL operators (and their equivalents) avoid loading the content of large strings and BLOB values from disk.

PR #81, opened 2023-08-24
cgosqlite: import SQLite amalgamation v3.43.0

crawshaw:


cgosqlite/shell.c.disabled
```diff
@@ -246,6 +246,7 @@ typedef unsigned char u8;
 #if SQLITE_OS_WINRT
 #include <intrin.h>
 #endif
+#undef WIN32_LEAN_AND_MEAN
```
creachadair, line 249: ?

creachadair: oh this was all from the import, nm

creachadair (approved):

crawshaw:
I ran `go test ./...` locally. We could put that in the CI.

The bigger set of tests are when the go.mod change is imported into corp, and we see if semantics have changed with our main set of tests.

PR #82, opened 2023-08-24
.github: add tests to CI

crawshaw:


.github/workflows/test-sqlite.yml
```diff
@@ -0,0 +1,24 @@
+name: test-sqlite
+
+on:
+  push:
+    branches: [ $default-branch ]
```
creachadair, line 5: Does this work here? I thought this was only for templates.

Do we need `${{ github.event.repository.default_branch }}` ?

crawshaw: ooh it worked this time!

lmk if you have a shorter form of this.

.github/workflows/test-sqlite.yml
```diff
@@ -0,0 +1,24 @@
+name: test-sqlite
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
```
creachadair, line 7: Consider maybe also
```yml
types: [opened, reopened, synchronize]
```

crawshaw: what does that do? (it seems to re-run on pushing to a PR already)

.github/workflows/test-sqlite.yml
creachadair: Yeah, I mean I just put in the specific name. It's not like default branch names change all that often anyway.

.github/workflows/test-sqlite.yml
creachadair: To be more specific:
```yml
on:
  push:
    branches:
      - main
  pull_request:
    types: [opened, reopened, synchronize]
```

(syntax mutatis mutandis)

.github/workflows/test-sqlite.yml
```diff
@@ -0,0 +1,24 @@
...
+
```
creachadair, line 8:
```suggestion
permissions:
  contents: read

```

though it is possible that is the default now.

crawshaw: I think it is. Going to skip this on the principle that fewer lines is better.

creachadair: It looks like the default is `read-all`, which is more general (e.g., it allows access to readable properties of the repo besides the files), but that is probably not a big issue.

.github/workflows/test-sqlite.yml
```diff
@@ -0,0 +1,24 @@
...
+    - name: Set up Go
+      uses: actions/setup-go@v4
+      with:
+        go-version: '1.21'
+
+    - name: Build
+      run: go build -v ./...
+
+    - name: Test
+      run: go test -v ./...
```
creachadair, line 24: `-race` ? 😄

creachadair (approved):

raggi:
@creachadair i bumped this, added race as you suggested, LGTY?

creachadair (approved):
I had forgotten about the existence of this PR, but yes, still LGTM. 😆

PR #85, opened 2023-11-23
sqlite: add an optional explicit interrupt on context termination

creachadair:
By default, a query with a time-consuming step may ignore termination of the
context passed in by the driver. Add an option allowing the caller to stipulate
that such queries should be explicitly interrupted (vi sqlite3_interrupt) when
the context ends. This option is disabled by default.

### DIFF
--- a/sqlite.go
+++ b/sqlite.go
@@ -490,6 +490,14 @@ func (s *stmt) ExecContext(ctx context.Context, args []driver.NamedValue) (drive
 	if err := s.bindAll(args); err != nil {
 		return nil, s.reserr("Stmt.Exec(Bind)", err)
 	}
+	if ctx.Value(queryCancelKey{}) != nil {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithCancel(ctx)
+		defer cancel()
+
+		db := s.stmt.DBHandle()
+		go func() { <-ctx.Done(); db.Interrupt() }()
+	}
 	row, lastInsertRowID, changes, duration, err := s.stmt.StepResult()
 	s.bound = false // StepResult resets the query
 	err = s.reserr("Stmt.Exec", err)
@@ -540,7 +548,13 @@ func (s *stmt) QueryContext(ctx context.Context, args []driver.NamedValue) (driv
 	if err := s.bindAll(args); err != nil {
 		return nil, err
 	}
-	return &rows{stmt: s}, nil
+	cancel := func() {}
+	if ctx.Value(queryCancelKey{}) != nil {
+		ctx, cancel = context.WithCancel(ctx)
+		db := s.stmt.DBHandle()
+		go func() { <-ctx.Done(); db.Interrupt() }()
+	}
+	return &rows{stmt: s, cancel: cancel}, nil
 }
 
 func (s *stmt) resetAndClear() error {
@@ -717,6 +731,7 @@ func colDeclTypeFromString(s string) colDeclType {
 type rows struct {
 	stmt   *stmt
 	closed bool
+	cancel context.CancelFunc // call when query ends
 
 	// colType is the column types for Step to fill on each row. We only use 23
 	// as it packs well with the closed bool byte above (24 bytes total, same as
@@ -764,6 +779,7 @@ func (r *rows) Close() error {
 		return ErrClosed
 	}
 	r.closed = true
+	r.cancel()
 	if err := r.stmt.resetAndClear(); err != nil {
 		return r.stmt.reserr("Rows.Close(Reset)", err)
 	}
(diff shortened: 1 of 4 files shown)


sqlite.go
```diff
@@ -490,6 +490,11 @@ func (s *stmt) ExecContext(ctx context.Context, args []driver.NamedValue) (drive
 	if err := s.bindAll(args); err != nil {
 		return nil, s.reserr("Stmt.Exec(Bind)", err)
 	}
+	if d, ok := ctx.Deadline(); ok && ctx.Value(queryTimeoutKey{}) != nil {
```
crawshaw, line 493: This is basically what I was thinking, but you can handle general context cancelation here, not just deadlines, by using a goroutine and selecting on ctx.Done

creachadair: I did this so we wouldn't need another goroutine unless the cancellation actually fires.

crawshaw: If `time.AfterFunc` is really that efficient, maybe it can always be on?

For most queries most of the time, spinning up a goroutine during the query is totally fine. It's just in our server full of hot paths we can't do it. So the idea of having an opt-in way to make contexts fully-functional seems very reasonable to me.

creachadair: AfterFunc still starts a goroutine when the timer ends, but in the common case where a query doesn't exceed its deadline it's just the timer entry. But if you don't think a goroutine for all queries is a concern that's just as easy to wire up. I'll do that.

creachadair: I pushed a separate commit to watch the context for any termination as you suggested.

sqlite.go
```diff
@@ -490,6 +490,14 @@ func (s *stmt) ExecContext(ctx context.Context, args []driver.NamedValue) (drive
 	if err := s.bindAll(args); err != nil {
 		return nil, s.reserr("Stmt.Exec(Bind)", err)
 	}
+	if ctx.Value(queryCancelKey{}) != nil {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithCancel(ctx)
+		defer cancel()
+
+		db := s.stmt.DBHandle()
+		go func() { <-ctx.Done(); db.Interrupt() }()
```
crawshaw, line 499: It might be worth only calling `Interrupt` if the statement is still running, which you could check with `s.bound` I believe.

sqlite.go
```diff
@@ -540,7 +548,13 @@ func (s *stmt) QueryContext(ctx context.Context, args []driver.NamedValue) (driv
 	if err := s.bindAll(args); err != nil {
 		return nil, err
 	}
-	return &rows{stmt: s}, nil
+	cancel := func() {}
+	if ctx.Value(queryCancelKey{}) != nil {
+		ctx, cancel = context.WithCancel(ctx)
+		db := s.stmt.DBHandle()
+		go func() { <-ctx.Done(); db.Interrupt() }()
```
crawshaw, line 555: similarly here I think if you define `*rows` before the if statement, you could condition calling Interrupt on `rows.closed`

crawshaw (approved):

sqlite.go
creachadair: I think if I do that I'll have to add a lock or an atomic on closed, which would otherwise become racy.

crawshaw:
I think it's fine to spin up a goroutine here because the only time this context value is going to be used is when the query is either known to be slow or its runtime is extremely variable, and the runtime is user-visible.

There's never going to be a query like that in a hot path of a very busy server because it makes for an impossible to manage service. But in cases like your use case, the overhead of a goroutine per query is ~0.

sqlite.go
crawshaw: oof yeah you're right.

taking a quick read of the sqlite3 sources, sqlite3_interrupt is an atomic store, that is unconditionally stored over when _step is called. so 99% of the cost of unconditionally calling interrupt is just the cgo overhead. given this path adds a goroutine, it's not worth optimizing. so this is fine as is.

creachadair:
@crawshaw do you have suggestions for good ways to test this? I guess I could just make a really big single-row recursive query on a counter, but that seems like an invitation for flakes. I thought about injecting a user-defined function, though to do that I'd have to add some extra mass to expose those parts of the library. Do you have other clever ideas?

crawshaw:
I think you can make a CTE that effectively runs forever, to avoid flakes:

```
WITH RECURSIVE inf(n) AS (
    SELECT 1
    UNION ALL
    SELECT n+1 FROM inf
)
SELECT * FROM inf;
```

creachadair:
Ah! I didn't realize inf worked there. I'll use that, thanks!

creachadair:
PTAL, now with test. 😄

sqlite.go
creachadair: (same issue as noted below)

sqlite_test.go
```diff
@@ -398,6 +398,45 @@ func TestWithPersist(t *testing.T) {
...
+			t.Error("Rows did not report an error")
+		} else if !strings.Contains(err.Error(), "SQLITE_INTERRUPT") {
+			t.Errorf("Rows err=%v, want SQLITE_INTERRUPT", err)
+		}
+	}()
+
+	select {
+	case <-done:
+		// OK
+	case <-time.After(time.Second):
```
crawshaw, line 435: I'd suggest bumping this because Github CI is an Amazing Place.

crawshaw (approved):
Very nice, thanks!

PR #87, opened 2023-11-28
sqlite: fix handling of "early" context cancellations

creachadair:
The optional cancellation hook in #85 and #86 has a deficiency:
If the query successfully completes before its context ends, the cleanup
context would unconditionally trigger an interrupt on the database connection.
That interrupt could race with a subsequent query on that connection, and cause
a spurious cancellation.

To fix this, separate the cleanup context from the input context, and only
effect an interrupt if the _input_ context terminates before cleanup occurs.
Only if cleanup is definitively prior to the query finishing will we effect an
explicit interrupt.

Also: Add a test that demonstrates the original problem.

### DIFF
--- a/sqlite.go
+++ b/sqlite.go
@@ -491,12 +491,23 @@ func (s *stmt) ExecContext(ctx context.Context, args []driver.NamedValue) (drive
 		return nil, s.reserr("Stmt.Exec(Bind)", err)
 	}
 	if ctx.Value(queryCancelKey{}) != nil {
-		var cancel context.CancelFunc
-		ctx, cancel = context.WithCancel(ctx)
-		defer cancel()
+		pctx, pcancel := context.WithCancel(ctx)
+		defer pcancel()
 
 		db := s.stmt.DBHandle()
-		go func() { <-ctx.Done(); db.Interrupt() }()
+		go func() {
+			running := pctx.Err() == nil
+			select {
+			case <-pctx.Done():
+				// OK, we fell off the end of the exec
+			case <-ctx.Done():
+				// The input context ended; if we haven't already done a cleanup on
+				// this query, interrupt it.
+				if running {
+					db.Interrupt()
+				}
+			}
+		}()
 	}
 	row, lastInsertRowID, changes, duration, err := s.stmt.StepResult()
 	s.bound = false // StepResult resets the query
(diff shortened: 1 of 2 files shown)


sqlite.go
```diff
@@ -491,12 +491,23 @@ func (s *stmt) ExecContext(ctx context.Context, args []driver.NamedValue) (drive
...
 	if ctx.Value(queryCancelKey{}) != nil {
-		var cancel context.CancelFunc
-		ctx, cancel = context.WithCancel(ctx)
-		defer cancel()
+		pctx, pcancel := context.WithCancel(ctx)
+		defer pcancel()
 
 		db := s.stmt.DBHandle()
-		go func() { <-ctx.Done(); db.Interrupt() }()
+		go func() {
```
crawshaw, line 498: Something about this doesn't seem right, but I'm struggling to explain it clearly. See if this makes sense:

If ctx is done, then pctx is done. That means canceling ctx immediately makes both conditions in the select statement selectable. The nature of selects are that they're fair (from the Go spec: "If one or more of the communications can proceed, a single one that can proceed is chosen via a uniform pseudo-random selection"). That means you only evaluate `running` half the time.

Even in the case where pctx is done first, there is an arbitrary delay until the select gets processed. It's possible to run `defer pcancel()` then later have ctx be done, and still you evaluate the ctx.Done() path, which reports running==true.

So this logic doesn't seem right.

creachadair: Another, maybe simpler framing, is that we only care to interrupt a query if its caller context (ctx) has ended before the query itself did. That can happen only if the query ran beyond its bounds. In that case we are certain that ctx is done, pctx is not.

But as you noted, the other way ctx can be done is if the query finished and returned control to the caller, ending ctx, and we happen to see that branch of the select by runtime choice. 

Then there are as I noted two possibilities: either that all happened before the goroutine even ran (outcome: no interrupt needed) or it overlapped with the execution of the query. We cannot be certain it's STILL running, but that's ok. If we wanted we could check pctx a second time, but I don't think we need to (though that might make it easier for the reader to see the brackets on the point of linearization of the interrupt)

creachadair: I reworked the plumbing a bit in a way that I think should address your concerns (it doesn't change the semantics, but should avoid some of the confusing edges).

Notably, it is _still_ possible that the scheduler doesn't wake up the recv on pctx.Done() until "much later", in which case we could interrupt the db during a subsequent query. That we can't entirely prevent at the query level—we'd have to bake some machinery into the conn so that it couldn't be returned to the active set before the goroutine finishes. 

But the more immediate issue here is avoiding the race with _our own cancellation_.

creachadair: Another idea in 0ce946dae, less invasive than decorating the whole conn.

PR #90, opened 2023-12-16
sqlite: revert context-based cancellation hooks from #85

creachadair:
The interaction between cleanup and reuse of the connection makes this feature
hard to use as-written. While #87 proposes a couple ways to fix the issue,
let's revert the partial state of things; we can fold these changes in with the
fix once we have consensus.

- Revert "sqlite: cancel completed queries after cleanup, not before (#86)"
  This reverts commit 78a5f32445416db077d122abf09e8aa78202246e.

- Revert "sqlite: add an optional explicit interrupt on context termination (#85)"
  This reverts commit 607d0c19ac5c651c2a4f9f9836552ef9be08221d.

### DIFF
--- a/cgosqlite/cgosqlite.go
+++ b/cgosqlite/cgosqlite.go
@@ -132,10 +132,6 @@ func (db *DB) Close() error {
 	return errCode(res)
 }
 
-func (db *DB) Interrupt() {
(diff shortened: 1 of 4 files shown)


creachadair:
@crawshaw this one should not require detailed review 🙂

PR #91, opened 2024-01-09
sqlstats: export Collect for metrics in tests

kradalby:
This PR exports the QueryStats and Collect method so it can be used to gather statistics and used during tests.

Used in https://github.com/tailscale/corp/pull/16678

### DIFF
--- a/sqlstats/sqlstats.go
+++ b/sqlstats/sqlstats.go
@@ -38,7 +38,21 @@ type Tracer struct {
 	// in the steady state, a sync.Map would be a faster object
 	// here.
 	mu      sync.RWMutex
-	queries map[string]*queryStats // normalized query -> stats
+	queries map[string]*QueryStats // normalized query -> stats
+}
+
+// Reset resets the state of t to its initial conditions.
+func (t *Tracer) Reset() {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+
+	t.TxCount = nil
+	t.TxCommit = nil
+	t.TxCommitError = nil
+	t.TxRollback = nil
+	t.TxTotalSeconds = nil
+	t.curTxs = sync.Map{}
+	t.queries = nil
 }
 
 type connStats struct {
@@ -70,14 +84,26 @@ func (t *Tracer) done(s *connStats) (why string, readOnly bool) {
 	return why, readOnly
 }
 
-type queryStats struct {
-	query string
+// QueryStats is a collection of stats for a given Query.
+type QueryStats struct {
+	Query string
 
 	// When inside the queries map all fields must be accessed as atomics.
-	count    int64
-	errors   int64
-	duration int64 // time.Duration
-	mean     int64
+
+	// Count represents the number of times this query has been
+	// executed.
+	Count int64
+
+	// Errors represents the number of errors encountered executing
+	// this query.
+	Errors int64
+
+	// TotalDuration represents the accumulated time spent executing the query.
+	TotalDuration time.Duration
+
+	// MeanDuration represents the average time spent executing the query.
+	MeanDuration time.Duration
+
 	// TODO lastErr atomic.Value
 }
 
@@ -90,7 +116,7 @@ func normalizeQuery(q string) string {
 	return q
 }
 
-func (t *Tracer) queryStats(query string) *queryStats {
+func (t *Tracer) queryStats(query string) *QueryStats {
 	query = normalizeQuery(query)
 
 	t.mu.RLock()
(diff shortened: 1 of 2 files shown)


sqlstats/sqlstats.go
```diff
@@ -70,14 +83,14 @@ func (t *Tracer) done(s *connStats) (why string, readOnly bool) {
 	return why, readOnly
 }
 
-type queryStats struct {
-	query string
+type QueryStats struct {
```
creachadair, line 86: This should get a doc comment now that it's exported.

sqlstats/sqlstats.go
```diff
@@ -70,14 +83,14 @@ func (t *Tracer) done(s *connStats) (why string, readOnly bool) {
...
+	Query string
 
 	// When inside the queries map all fields must be accessed as atomics.
-	count    int64
-	errors   int64
-	duration int64 // time.Duration
-	mean     int64
+	Count    int64
+	Errors   int64
+	Duration int64 // time.Duration
```
creachadair, line 92: Does the comment mean we should convert it into a `time.Duration` or that the units are nanoseconds? (Or both?) Unless the extra cost of conversion is a performance consideration, I'd suggest making it a duration anyway—and if we don't want to do that for some reason we should probably put the units in the name, e.g., `DurationMicros` or something.

sqlstats/sqlstats.go
```diff
@@ -104,28 +117,29 @@ func (t *Tracer) queryStats(query string) *queryStats {
...
 	if stats == nil {
-		stats = &queryStats{query: query}
+		stats = &QueryStats{Query: query}
 		t.queries[query] = stats
 	}
 	return stats
 }
 
-func (t *Tracer) collect() (rows []queryStats) {
+func (t *Tracer) Collect() (rows []QueryStats) {
```
creachadair, line 130: Since everywhere else we pass `QueryStats` by its pointer, we should probably do that here too. `[]*QueryStats`

(Alternatively we could switch it to pass by value everywhere, but it's a fairly big struct so a pointer is fine).

sqlstats/sqlstats.go
```diff
@@ -70,14 +83,14 @@ func (t *Tracer) done(s *connStats) (why string, readOnly bool) {
...
 
 	// When inside the queries map all fields must be accessed as atomics.
-	count    int64
-	errors   int64
-	duration int64 // time.Duration
-	mean     int64
+	Count    int64
+	Errors   int64
+	Duration int64 // time.Duration
+	Mean     int64
```
creachadair, line 93: Please document the units here too.

sqlstats/sqlstats.go
```diff
@@ -70,14 +83,14 @@ func (t *Tracer) done(s *connStats) (why string, readOnly bool) {
...
-	query string
+type QueryStats struct {
+	Query string
 
 	// When inside the queries map all fields must be accessed as atomics.
-	count    int64
-	errors   int64
-	duration int64 // time.Duration
-	mean     int64
+	Count    int64
```
creachadair, line 90: Is this row count?

creachadair (commented):
The semantics LGTM, I have a few requests around documentation and API surface.

sqlstats/sqlstats.go
```diff
@@ -164,7 +165,15 @@ func (t *Tracer) Query(prepCtx context.Context, id sqliteh.TraceConnID, query st
...
 	atomic.AddInt64(&stats.Count, 1)
-	atomic.AddInt64(&stats.Duration, int64(duration))
+
+	newDuration := stats.Duration.Load()
+	if newDuration == nil {
+		newDuration = &duration
+	} else {
+		*newDuration = *newDuration + duration
+	}
+	stats.Duration.Store(newDuration)
```
icio, line 175: Separating the load and store means this is no-longer an atomic Add, like AddInt64 is. If you want to make `stats.Duration` a true time.Duration then you can just convert it in the eyes of the AddInt64 call: https://go.dev/play/p/1Xeiyf3l1Zt

```go
// If stats.Duration is a time.Duration you can do:
atomic.AddInt64((*int64)(&stats.Duration), int64(duration))
```

And then you won't need to call Load() everywhere else.

sqlstats/sqlstats.go
```diff
@@ -70,14 +83,27 @@ func (t *Tracer) done(s *connStats) (why string, readOnly bool) {
...
-	count    int64
-	errors   int64
-	duration int64 // time.Duration
-	mean     int64
+
+	// Count represents the number of times this query has been
+	// executed during a trace.
+	Count int64
+
+	// Errors represents the number of error encountered executing
```
icio, line 96:
```suggestion
	// Errors represents the number of errors encountered executing
```

sqlstats/sqlstats.go
```diff
@@ -70,14 +83,27 @@ func (t *Tracer) done(s *connStats) (why string, readOnly bool) {
...
+	Query string
 
 	// When inside the queries map all fields must be accessed as atomics.
-	count    int64
-	errors   int64
-	duration int64 // time.Duration
-	mean     int64
+
+	// Count represents the number of times this query has been
+	// executed during a trace.
```
icio, line 93: The tracer is passed to sqlite.Connector and used for the life of the database/sql.DB - "during a trace" feels a bit short-lived in that context.

sqlstats/sqlstats.go
```diff
@@ -70,14 +83,27 @@ func (t *Tracer) done(s *connStats) (why string, readOnly bool) {
...
+	// this query during a trace.
+	Errors int64
+
+	// Duration represents the accumulated time spent executing the query
+	// during a trace.
+	Duration time.Duration
+
+	// Mean represents the average time spent executing the query during a trace
+	// in nanoseconds.
+	Mean int64
```
icio, line 106: This could also be a time.Duration.

sqlstats/sqlstats.go
```diff
@@ -70,14 +83,27 @@ func (t *Tracer) done(s *connStats) (why string, readOnly bool) {
...
+	// executed during a trace.
+	Count int64
+
+	// Errors represents the number of error encountered executing
+	// this query during a trace.
+	Errors int64
+
+	// Duration represents the accumulated time spent executing the query
+	// during a trace.
+	Duration time.Duration
```
icio, line 102: I wonder if TotalDuration/MeanDuration would be clearer than Duration/Mean.

sqlstats/sqlstats.go
```diff
@@ -104,40 +130,45 @@ func (t *Tracer) queryStats(query string) *queryStats {
...
 	return rows
 }
 
 func (t *Tracer) Query(prepCtx context.Context, id sqliteh.TraceConnID, query string, duration time.Duration, err error) {
 	stats := t.queryStats(query)
 
-	atomic.AddInt64(&stats.count, 1)
-	atomic.AddInt64(&stats.duration, int64(duration))
+	atomic.AddInt64(&stats.Count, 1)
+	atomic.AddInt64((*int64)(&stats.Duration), int64(duration))
```
icio, line 168: 👍

icio (commented):
LGTM

sqlstats/sqlstats.go
```diff
@@ -38,7 +38,20 @@ type Tracer struct {
 	// in the steady state, a sync.Map would be a faster object
 	// here.
 	mu      sync.RWMutex
-	queries map[string]*queryStats // normalized query -> stats
+	queries map[string]*QueryStats // normalized query -> stats
+}
+
+func (t *Tracer) Reset() {
```
creachadair, line 44:
```go
// Reset resets the state of t to its initial conditions.
```

or words to that effect.

creachadair (approved):

PR #92, opened 2024-01-23
sqlstats: fix Tracer.Reset data races

icio:
I failed to test the previous Reset implementation with `-race`, thinking that because it was only used in a Test, a straightforward zeroing of all of the fields would suffice. This proved not to be the case, and so I'm fixing it now:

```
$ go test ./sqlstats/ -run Race -race -count 20
ok  	github.com/tailscale/sqlite/sqlstats	21.289s
```

### DIFF
--- a/sqlstats/sqlstats_test.go
+++ b/sqlstats/sqlstats_test.go
@@ -3,12 +3,14 @@ package sqlstats
 import (
 	"context"
 	"database/sql"
+	"fmt"
 	"io"
 	"net/http"
 	"net/http/httptest"
 	"slices"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/tailscale/sqlite"
 )
@@ -117,5 +119,66 @@ func TestCollect(t *testing.T) {
 		if query.Count != int64(wantCount[idx]) {
 			t.Errorf("unexpected query count for %q, got: %d, expected: %d", query.Query, query.Count, wantCount[idx])
 		}
+	}
+}
+
+func TestTracerResetRace(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+
+	tracer := &Tracer{}
+	db := sql.OpenDB(sqlite.Connector("file:"+t.TempDir()+"/test.db", nil, tracer))
+	defer db.Close()
+
+	tx, err := db.BeginTx(ctx, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer tx.Rollback()
+	if _, err := tx.ExecContext(ctx, "CREATE TABLE t (c);"); err != nil {
+		t.Fatal(err)
+	}
+	if err := tx.Commit(); err != nil {
+		t.Fatal(err)
+	}
+
+	// Continually reset the tracer.
+	go func() {
+		for ctx.Err() == nil {
+			tracer.Reset()
+		}
+	}()
+
+	// Continually grab transactions and insert into the database.
+	errc := make(chan error)
+	go func() {
+		defer close(errc)
+		var n int64
+		for ctx.Err() == nil {
+			n++
+			tx, err = db.BeginTx(ctx, nil)
+			if err != nil {
+				errc <- fmt.Errorf("starting tx: %w", err)
+				return
+			}
+			if _, err = tx.Exec("INSERT INTO t VALUES(?)", n); err != nil {
+				errc <- fmt.Errorf("insert: %w", err)
+				return
+			}
+			if n%2 == 0 {
+				if err := tx.Commit(); err != nil {
+					errc <- fmt.Errorf("commit: %w", err)
+					return
+				}
+			}
+			tx.Rollback()
+		}
+	}()
+
+	for err := range errc {
+		if ctx.Err() != nil {
+			return
+		}
+		t.Fatalf("unexpected error: %s", err)
 	}
 }
--- a/sqlstats/sqlstats.go
+++ b/sqlstats/sqlstats.go
@@ -43,15 +43,28 @@ type Tracer struct {
 
 // Reset resets the state of t to its initial conditions.
 func (t *Tracer) Reset() {
+	if t.TxCount != nil {
+		t.TxCount.Init()
+	}
+	if t.TxCommit != nil {
+		t.TxCommit.Init()
+	}
+	if t.TxCommitError != nil {
+		t.TxCommitError.Init()
+	}
+	if t.TxRollback != nil {
+		t.TxRollback.Init()
+	}
+	if t.TxTotalSeconds != nil {
+		t.TxTotalSeconds.Init()
+	}
+	t.curTxs.Range(func(key, value any) bool {
+		t.curTxs.Delete(key)
+		return true
+	})
+
 	t.mu.Lock()
 	defer t.mu.Unlock()
-
-	t.TxCount = nil
-	t.TxCommit = nil
-	t.TxCommitError = nil
-	t.TxRollback = nil
-	t.TxTotalSeconds = nil
-	t.curTxs = sync.Map{}
 	t.queries = nil
 }
 


icio:
Not sure if this error is related? https://github.com/tailscale/corp/actions/runs/7626448916/job/20772858826?pr=16701

icio:
Not sure if this error is related?

```
# github.com/tailscale/sqlite/cgosqlite
In file included from /nix/store/d3bkp4jq1n6kxvhgs45h5zc8ppg759cm-glibc-2.37-8-dev/include/string.h:535,
                 from sqlite3.c:14349:
In function ‘memset’,
    inlined from ‘sqlite3Fts5MallocZero’ at sqlite3.c:229740:7,
    inlined from ‘fts5IndexTombstoneRebuild’ at sqlite3.c:241450:35,
    inlined from ‘fts5IndexTombstoneAdd’ at sqlite3.c:241521:3,
    inlined from ‘sqlite3Fts5IndexContentlessDelete’ at sqlite3.c:241561:11,
    inlined from ‘fts5StorageContentlessDelete’ at sqlite3.c:246382:10,
    inlined from ‘sqlite3Fts5StorageDelete’ at sqlite3.c:246486:12:
/nix/store/d3bkp4jq1n6kxvhgs45h5zc8ppg759cm-glibc-2.37-8-dev/include/bits/string_fortified.h:59:10: warning: ‘__builtin_memset’ specified size between [92](https://github.com/tailscale/corp/actions/runs/7626448916/job/20772858826#step:5:93)23372036854775808 and 1844674407370[95](https://github.com/tailscale/corp/actions/runs/7626448916/job/20772858826#step:5:96)51615 exceeds maximum object size 9223372036854775807 [-Wstringop-overflow=]
   59 |   return __builtin___memset_chk (__dest, __ch, __len,
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   60 |                                  __glibc_objsize0 (__dest));
      |                                  ~~~~~~~~~~~~~~~~~~~~~~~~~~
```

from https://github.com/tailscale/corp/actions/runs/7626448916/job/20772858826

sqlstats/sqlstats_test.go
```diff
@@ -119,3 +122,90 @@ func TestCollect(t *testing.T) {
 		}
 	}
 }
+
+func TestTracerResetRace(t *testing.T) {
```
creachadair, line 126: I believe this test can be simplified a bit to make it easier to read. Specifically:

- We don't need to cancel conditionally; cancelling a completed context is already idempotent.
- We don't need to count error values, only the transaction goroutine reports any errors that need to be reported.
- We can simplify the context-done checks a bit.

```go
func TestTracerResetRace(t *testing.T) {
	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
	defer cancel()

	tracer := &Tracer{}
	db := sql.OpenDB(sqlite.Connector("file:"+t.TempDir()+"/test.db", nil, tracer))
	defer db.Close()

	if _, err := db.ExecContext(ctx, "CREATE TABLE t (c);"); err != nil {
		t.Fatalf("Create table: %v", err)
	}

	errc := make(chan error)
	go func() {
		defer close(errc)
		defer func() {
			if r := recover(); r != nil {
				ch <- fmt.Errorf("panic: %v", r)
			}
		}()
		for n := 0; ctx.Err() == nil; n++ {
			tx, err := db.BeginTx(ctx, nil)
			if err != nil {
				ch <- err
				return
			} else if _, err := tx.Exec("INSERT INTO t VALUES (?)", n); err != nil {
				tx.Rollback()
				ch <- err
				return
			} else if n%2 == 0 {
				continue
			}
			if err := tx.Commit(); err != nil {
				ch <- err
				return
			}
		}

	}()
	go func() {
		for ctx.Err() == nil {
			tracer.Reset()
		}
	}()

	for err := range errc {
		t.Errorf("Unexpected transaction error: %v", err)
		cancel()
	}
}

```

sqlstats/sqlstats_test.go
```diff
@@ -119,3 +122,90 @@ func TestCollect(t *testing.T) {
...
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	maybeCancel := func() {
```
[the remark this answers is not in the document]
icio: Ahha! I thought it panicked like `close(chan)`. Thanks for pointing this out.

sqlstats/sqlstats_test.go
icio: Thanks for those pointers, that's much simpler. My original code certainly does look like a bit of an overreaction to not having errgroup to reach for 😅

creachadair: I almost suggested importing errgroup, but then I realized we probably don't really need it because one of the two goroutines never reports anything 🙂

sqlstats/sqlstats_test.go
```diff
@@ -119,3 +121,62 @@ func TestCollect(t *testing.T) {
...
+
+	// Continually reset the tracer.
+	go func() {
+		for ctx.Err() == nil {
+			tracer.Reset()
+		}
+	}()
+
+	// Continually grab transactions and insert into the database.
+	errc := make(chan error, 1)
```
creachadair, line 153: We shouldn't need a buffer here, does the test not work without it? (If so I'd like to understand why, since AFAIK it should be fine)

creachadair (approved):

sqlstats/sqlstats_test.go
```diff
@@ -119,3 +121,62 @@ func TestCollect(t *testing.T) {
...
+				if err := tx.Commit(); err != nil {
+					errc <- fmt.Errorf("commit: %w", err)
+					return
+				}
+			}
+			tx.Rollback()
+		}
+	}()
+
+	err = <-errc
```
creachadair, line 178: Is there some reason not to loop over the channel? It is possible multiple errors could be reported, and we still want all the goroutines to be able to exit.

sqlstats/sqlstats_test.go
```diff
@@ -119,3 +121,62 @@ func TestCollect(t *testing.T) {
...
+	go func() {
+		for ctx.Err() == nil {
+			tracer.Reset()
+		}
+	}()
+
+	// Continually grab transactions and insert into the database.
+	errc := make(chan error, 1)
+	go func() {
+		defer close(errc)
```
creachadair, line 155: Was the panic check not desired?

icio: Not undesirable - just unnecessary, as the test will fail there's a panic in any of these goroutines. I added it originally because for some reason I thought the default behaviour was to let the goroutine die quietly, but it crashes the whole program.

sqlstats/sqlstats_test.go
icio: There's only one goroutine writing to it now. Originally I was thinking that'd want to receive an error and ensure it's nil, but I didn't actually implement that behaviour - the channel is just closed with no sent errors when everything works successfully. I've switched to a loop.

sqlstats/sqlstats_test.go
icio: Buffer not required - I've removed it.

sqlstats/sqlstats_test.go
creachadair: I was originally thinking of the panic handler which could write a second value, but now that's done. 😀

creachadair:
> Not sure if this error is related?
> 
> ```
> # github.com/tailscale/sqlite/cgosqlite
> In file included from /nix/store/d3bkp4jq1n6kxvhgs45h5zc8ppg759cm-glibc-2.37-8-dev/include/string.h:535,
>                  from sqlite3.c:14349:
> In function ‘memset’,
>     inlined from ‘sqlite3Fts5MallocZero’ at sqlite3.c:229740:7,
>     inlined from ‘fts5IndexTombstoneRebuild’ at sqlite3.c:241450:35,
>     inlined from ‘fts5IndexTombstoneAdd’ at sqlite3.c:241521:3,
>     inlined from ‘sqlite3Fts5IndexContentlessDelete’ at sqlite3.c:241561:11,
>     inlined from ‘fts5StorageContentlessDelete’ at sqlite3.c:246382:10,
>     inlined from ‘sqlite3Fts5StorageDelete’ at sqlite3.c:246486:12:
> /nix/store/d3bkp4jq1n6kxvhgs45h5zc8ppg759cm-glibc-2.37-8-dev/include/bits/string_fortified.h:59:10: warning: ‘__builtin_memset’ specified size between [92](https://github.com/tailscale/corp/actions/runs/7626448916/job/20772858826#step:5:93)23372036854775808 and 1844674407370[95](https://github.com/tailscale/corp/actions/runs/7626448916/job/20772858826#step:5:96)51615 exceeds maximum object size 9223372036854775807 [-Wstringop-overflow=]
>    59 |   return __builtin___memset_chk (__dest, __ch, __len,
>       |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
>    60 |                                  __glibc_objsize0 (__dest));
>       |                                  ~~~~~~~~~~~~~~~~~~~~~~~~~~
> ```
> 
> from https://github.com/tailscale/corp/actions/runs/7626448916/job/20772858826

I do not think this is related in any way to your change here. It's a valid complaint from the compiler about what appears to be a size range issue ([originating here](https://github.com/tailscale/sqlite/blob/main/cgosqlite/sqlite3.c#L246486)) but I don't see any way your change could have triggered this. It's from cleanup code in SQLite itself, and the changes in this PR do not touch SQLite data structures.

This _may_ mean we want to update our version of the sqlite library itself, though I'd do that separately from this change if we do.

icio:
Thanks for confirming!

PR #96, opened 2024-05-14
Fix macOS build regression from XOPEN_SOURCE=600

raggi:
This fixes the regression, also adds a macOS CI pipeline.


creachadair (approved):
Oops, sorry, I should have thought to ask about Darwin.

raggi:
It's easy to miss, libc/posix versioning are not pleasant to deal with!

PR #101, opened 2024-05-24
cgosqlite: cleanup the import layout some more

raggi:
Just removes duplicate imports and orders them better.

### DIFF
--- a/cgosqlite/cgosqlite.go
+++ b/cgosqlite/cgosqlite.go
@@ -33,21 +33,10 @@ package cgosqlite
 // // Select POSIX 2014 at least for clock_gettime.
 // #cgo CFLAGS: -D_XOPEN_SOURCE=600
 // #cgo CFLAGS: -D_DARWIN_C_SOURCE=1
-// #cgo linux CFLAGS: -std=c99
 //
-// // On Android, unlike Linux, there are no separate libpthread or librt
-// // libraries. That functionality is included directly in libc, which does not
-// // need to be explicitly linked against. See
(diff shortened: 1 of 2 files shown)


cgosqlite/cgosqlite.h
```diff
@@ -1,4 +1,6 @@
-#include <stdint.h>
```
creachadair, line 1: I think we'll still need `<stdint.h>` here.

cgosqlite/cgosqlite.go
```diff
@@ -33,21 +33,7 @@ package cgosqlite
 // // Select POSIX 2014 at least for clock_gettime.
 // #cgo CFLAGS: -D_XOPEN_SOURCE=600
 // #cgo CFLAGS: -D_DARWIN_C_SOURCE=1
-// #cgo linux CFLAGS: -std=c99
 //
-// // On Android, unlike Linux, there are no separate libpthread or librt
```
creachadair, line 38: Some of the imports are redundant, but are we sure that the other compiler directives are too? I'm not sure the tests in this repo are exercising that.

raggi: we pass `-std=gnu11` in gocross, the `-std=c99` is somewhat redundant. sqlite itself still builds with c89 and only requires a few memory functions and an allocator, but turns out that fts5 is the thing that depends on libm.

raggi: (but on clang systems, log(), which it needs is provided without needing to link anything)

creachadair (approved):

PR #102, opened 2024-05-24
cgosqlite: set SQLITE_TEMP_STORE=1

raggi:
SQLITE_TEMP_STORE=1 allows PRAGMA temp_store = MEMORY, while keeping temporary files as the default behavior for temporary storage.

This can be used to accelerate some migrations, when building an in-memory temporary btree to use in a later correlated subquery is faster than the query planner would normally do things, for example when optimizing locality or btree operation order.


bradfitz:
What's with the compilation error?

raggi:
> What's with the compilation error?

It's on top of another patch, but didn't need to be, and is no longer.

bradfitz (approved):

PR #103, opened 2024-07-11
sqlite: fix handling of "early" context cancellations

creachadair:
(A restatement of #87)

The optional cancellation hook in #85 and #86 has a deficiency: If the query successfully completes before its context ends, the cleanup context would unconditionally trigger an interrupt on the database connection. That interrupt could race with a subsequent query on that connection, and cause a spurious cancellation.

To fix this, separate the cleanup context from the input context, and only effect an interrupt if the _input_ context terminates before cleanup occurs. Only if cleanup is definitively prior to the query finishing will we effect an explicit interrupt.

Also: Add a test that demonstrates the original problem.

### DIFF
--- a/sqlite.go
+++ b/sqlite.go
@@ -494,7 +494,7 @@ func (s *stmt) ExecContext(ctx context.Context, args []driver.NamedValue) (drive
 	if ctx.Value(queryCancelKey{}) != nil {
 		done = make(chan struct{})
 		pctx, pcancel := context.WithCancel(ctx)
-		defer pcancel()
+		defer pcancel() // to make the AfterFunc fire and close(done)
 
 		db := s.stmt.DBHandle()
 		stop := context.AfterFunc(pctx, func() {
--- a/sqlite.go
+++ b/sqlite.go
@@ -756,7 +756,7 @@ type rows struct {
 	stmt   *stmt
 	closed bool
 	cancel context.CancelFunc // call when query ends
-	done   chan struct{}      // closed when cancellation is done
+	done   chan struct{}      // either nil, or closed when cancellation is done
 
 	// colType is the column types for Step to fill on each row. We only use 23
 	// as it packs well with the closed bool byte above (24 bytes total, same as
--- a/sqlite.go
+++ b/sqlite.go
@@ -490,13 +490,22 @@ func (s *stmt) ExecContext(ctx context.Context, args []driver.NamedValue) (drive
 	if err := s.bindAll(args); err != nil {
 		return nil, s.reserr("Stmt.Exec(Bind)", err)
 	}
+	done := make(chan struct{})
 	if ctx.Value(queryCancelKey{}) != nil {
-		var cancel context.CancelFunc
-		ctx, cancel = context.WithCancel(ctx)
-		defer cancel()
+		pctx, pcancel := context.WithCancel(ctx)
+		defer pcancel()
 
 		db := s.stmt.DBHandle()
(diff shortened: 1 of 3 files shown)


creachadair:
Ok, I figured out what I was doing wrong, and I think this is ready for a look.

sqlite.go
```diff
@@ -508,6 +518,7 @@ func (s *stmt) ExecContext(ctx context.Context, args []driver.NamedValue) (drive
 		return nil, err
 	}
 	_ = row // TODO: return error if exec on query which returns rows?
+	<-done
```
bradfitz, line 521: maybe do this conditionally on it being non-nil to avoid the alloc+close of it in the no-context-key case?

sqlite.go
```diff
@@ -490,13 +490,23 @@ func (s *stmt) ExecContext(ctx context.Context, args []driver.NamedValue) (drive
...
 
 		db := s.stmt.DBHandle()
-		go func() { <-ctx.Done(); db.Interrupt() }()
+		stop := context.AfterFunc(pctx, func() {
+			defer close(done)
+			if ctx.Err() != nil {
+				db.Interrupt()
+			}
+		})
+		// In the event we get an error from the initial step and exit early,
```
bradfitz, line 505: s/step/StepResult/ to clarify that "step" here means the sqlite3_step and not "initial step" in the general sense? (took me a second to parse it correctly)

But don't we want to stop this regardless, even in the ... oh! The normal flow is "defer pcancel()" to make the AfterFunc fire, which then does the close(done), when 521 blocks on. Might be worth a little comment like:

     defer pcancel() // to make the AfterFunc fire and close(done)

sqlite.go
```diff
@@ -732,6 +755,7 @@ type rows struct {
 	stmt   *stmt
 	closed bool
 	cancel context.CancelFunc // call when query ends
+	done   chan struct{}      // closed when cancellation is done
```
bradfitz, line 758: not exactly true in the no-context-key case?  then it's closed immediately.

But could also make this nil to mean no need to wait, like above?

bradfitz (approved):

bradfitz:
Bump `go.mod` from `go 1.20` to `go 1.21` for this?

sqlite.go
creachadair: Done, and so-documented.

sqlite.go
creachadair: Good idea, thanks.

sqlite.go
creachadair: SGTM, done.

creachadair:
The updates from review are in separate commits, but I'll squash them later.

sqlite.go
```diff
@@ -490,14 +490,24 @@ func (s *stmt) ExecContext(ctx context.Context, args []driver.NamedValue) (drive
...
 
 		db := s.stmt.DBHandle()
-		go func() { <-ctx.Done(); db.Interrupt() }()
+		stop := context.AfterFunc(pctx, func() {
+			defer close(done)
+			if ctx.Err() != nil {
+				db.Interrupt()
+			}
+		})
+		// In the event we get an error from the initial step and exit early,
```
bradfitz, line 506: still gonna clarify "initial step" into something that sounds more sqlite-specific?

bradfitz (approved):

sqlite.go
creachadair: Oops, yes, sorry I missed this one.

dsnet (approved):
Cool.

PR #104, opened 2024-07-12
sqlite: fix cancellation cleanup for ExecContext

creachadair:
I did not test the changes in #103 sufficiently, and missed a corner case in
ExecContext: We need to make sure the context cleanup happens prior to the
synchronization, not in a defer.

Add a test to actually trigger this case, and fix the cleanup so it happens
before return as intended.

### DIFF
--- a/sqlite.go
+++ b/sqlite.go
@@ -491,10 +491,6 @@ func (s *stmt) ExecContext(ctx context.Context, args []driver.NamedValue) (drive
 		return nil, s.reserr("Stmt.Exec(Bind)", err)
 	}
 
-	// If non-nil, wait must be called prior to returning to ensure a
-	// cancellation context (if present) has completed, so that a cancellation
-	// cannot outlast this request and fire during a later execution.
-	var wait func()
 	if ctx.Value(queryCancelKey{}) != nil {
 		done := make(chan struct{})
 		pctx, pcancel := context.WithCancel(ctx)
@@ -508,7 +504,11 @@ func (s *stmt) ExecContext(ctx context.Context, args []driver.NamedValue) (drive
 				db.Interrupt()
 			}
 		})
-		wait = func() { pcancel(); <-done }
+
+		// We must wait prior to returning to ensure a cancellation context (if
+		// present) has completed, so that a cancellation cannot outlast this
+		// request and fire during a later execution.
+		defer func() { pcancel(); <-done }()
 	}
 
 	row, lastInsertRowID, changes, duration, err := s.stmt.StepResult()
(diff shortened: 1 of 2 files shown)


sqlite.go
```diff
@@ -490,23 +490,22 @@ func (s *stmt) ExecContext(ctx context.Context, args []driver.NamedValue) (drive
...
 			defer close(done)
 			if ctx.Err() != nil {
 				db.Interrupt()
 			}
 		})
-		// In the event we get an error from the query's initial execution of
-		// sqlite3_step below and exit early, dissociate the cancellation since
-		// we don't want it to fire and potentially stop a later execution.
-		defer stop()
+		wait = func() { pcancel(); <-done }
```
maisem, line 508:
```
defer func() {
  pcancel()
  <-done
}()
```

would also work

sqlite.go
```diff
@@ -490,23 +490,22 @@ func (s *stmt) ExecContext(ctx context.Context, args []driver.NamedValue) (drive
...
-		done = make(chan struct{})
+		done := make(chan struct{})
 		pctx, pcancel := context.WithCancel(ctx)
-		defer pcancel() // to make the AfterFunc fire and close(done)
-
 		db := s.stmt.DBHandle()
-		stop := context.AfterFunc(pctx, func() {
+		context.AfterFunc(pctx, func() {
 			defer close(done)
 			if ctx.Err() != nil {
```
maisem, line 504: write a comment here why `ctx` and not `pctx`?

sqlite.go
maisem: or if you prefer
```
wait := func() { pcancel(); <-done }
defer wait()
```

maisem: fine as is too, just another approach

sqlite.go
creachadair: Good idea, done (also in the other spot we do this).

maisem (approved):
Thank you!

sqlite.go
creachadair: It won't quite work the way you describe it—if we bind the cleanup function in this scope, we can't call it below, and it's essential we do so prior to returning (since, as you noted in Slack, the defer doesn't happen till too late)

creachadair: Per our offline conversation: Actually in this case a defer is fine, because the code to check the results doesn't depend on the cancellation being resolved. So, I swapped it to a defer (since that avoids having to think about scopes at all). Thanks!

maisem (approved):

bradfitz (commented):
Nice. And more readable now too as a bonus.

